PR #11, opened 2023-06-30
Support for adding transaction in different currency for investments

### LINKED ISSUE #6: Different currencies error for investments
I added a transaction for the PHO stock account which sets the currency of the transaction to PHO:

```
USD|Buy PHO
USD|Buy PHO
USD|Buy PHO
USD|Buy PHO
USD|Dividend PHO
USD|Dividend PHO
USD|Dividend PHO
PHO|Sell PHO
```

This then causes an error when calculating totals as dinerojs doesn't know how to compute this. We should make the computations more clever as follows:

if the currency of the transaction is the same as the total's currency, then use that. If not, we should use the value/quantity instead (assuming that then that's the correct currency, if it's not it means indeed it is the wrong currency).

argaen:
This was harder than I expected and had to try different approaches.

Initially I tried to be able to parse the transactions regardless of the currency they are in but this added very messy code and complications and in the end I decided to just change the currency when adding the transaction. **This has the limitation that already existing transactions (when imported) will crash the app**.

The approach I decided consists on changing the currency when adding transactions from an investment account to be the fiat currency of the asset account and not the commodity of the investment. This gives consistency to investment transactions to always have the same transaction.

As an addition, there was a problem I didn't detect yet as never added a buy/sell transaction in maffin where we lack the mainCurrency exchange for that transaction. To fix this I added a new endpoint to Stocker that allows us to pull prices in the past for a given symbol. This endpoint had to appear for other future features so it's a good addition.

I've deleted some unused assets, unrelated to the PR.

Fixes #6


amplify/backend/function/stockerlambda/src/app.js
```diff
@@ -64,4 +64,30 @@ app.get('/api/prices/live', async (req, res) => {
   res.json(result);
 });
 
+app.get('/api/price', async (req, res) => {
```
argaen, line 67: This new endpoint allows us to retrieve historical prices for a given ticker. Will be useful for transaction form exchange rates too

src/book/lib/transaction/createTransaction.ts
```diff
@@ -0,0 +1,99 @@
...
+  // eslint-disable-next-line no-param-reassign
+  mainSplit.fk_transaction = transaction;
+
+  // We recalculate value for mainSplit in case we've swapped currencies
+  const { amount, scale } = toAmountWithScale(mainSplit.value * rate);
+  mainSplit.valueNum = amount;
+  mainSplit.valueDenom = parseInt('1'.padEnd(scale + 1, '0'), 10);
+
+  const mainCurrency = await getMainCurrency();
+  if (currency.guid !== mainCurrency.guid && isInvestment(mainSplit.account)) {
```
argaen, line 49: For investment transactions, we want to add a price instance to the DB so we know the value of the transaction in the main currency when representing it in the investments page.

PR #23, opened 2023-07-05
Populate exchange rate automatically

argaen:
When adding transactions, one of the most annoying things is to have to check exchange rates manually when the accounts have different currencies. Using the new endpoint added in #15 now we can load those rates dynamically.


src/components/forms/transaction/SplitField.tsx
```diff
@@ -40,6 +50,20 @@ export default function SplitField({
   if (splitCurrency && fromAccountCurrency !== splitCurrency) {
     showCurrencyExchange = true;
   }
+  const watchAccount = watch(`splits.${index}.toAccount`);
+
+  React.useEffect(() => {
```
argaen, line 55: Basically, fetch and update exchange rate if any of these change:

- split account
- date
- split currency or fromAccount currency

PR #24, opened 2023-07-05
Fix split order in transactions table

### LINKED ISSUE #5: Transactions table display wrong totals
Seems the ordering of splits is different when representing the relative amount vs the total calculated:

Screenshot 2023-06-28 at 10 54 20 AM

This seems to happen for transactions that have the same date

argaen:
Apparently react-table re-ordering was messing with the order of the items which came already ordered from typeorm query. What I've done is to disable the sorting from the table component and just rely on the ordering that comes from typeorm. Note that this means we can't order by date in the table anymore

Fixes #5

### DIFF
--- a/src/__tests__/app/user/logout/page.test.tsx
+++ b/src/__tests__/app/user/logout/page.test.tsx
@@ -1,7 +1,16 @@
 import React from 'react';
 import { render } from '@testing-library/react';
+import type { LinkProps } from 'next/link';
 
 import LogoutPage from '@/app/user/logout/page';
+
+jest.mock('next/link', () => jest.fn(
+  (
+    props: LinkProps & { children?: React.ReactNode } & React.HTMLAttributes<HTMLAnchorElement>,
+  ) => (
+    <a className={props.className} href={props.href.toString()}>{props.children}</a>
+  ),
+));
 
 describe('LogoutPage', () => {
   it('matches snapshot', () => {
--- a/src/__tests__/layout/LeftSideBar.test.tsx
+++ b/src/__tests__/layout/LeftSideBar.test.tsx
@@ -1,7 +1,16 @@
 import React from 'react';
 import { render } from '@testing-library/react';
+import type { LinkProps } from 'next/link';
 
 import LeftSidebar from '@/layout/LeftSidebar';
+
+jest.mock('next/link', () => jest.fn(
+  (
+    props: LinkProps & { children?: React.ReactNode } & React.HTMLAttributes<HTMLAnchorElement>,
+  ) => (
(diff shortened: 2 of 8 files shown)


src/__tests__/app/user/logout/page.test.tsx
```diff
@@ -1,8 +1,17 @@
...
 import LogoutPage from '@/app/user/logout/page';
 
+jest.mock('next/link', () => jest.fn(
+  (
+    props: LinkProps & { children?: React.ReactNode } & React.HTMLAttributes<HTMLAnchorElement>,
+  ) => (
+    <a className={props.className} href={props.href.toString()}>{props.children}</a>
+  ),
+));
+
```
argaen, line 14: unrelated but deletes a warning of missing `act` due to async update of `ForwardRef(LinkComponent)`.

src/__tests__/layout/LeftSideBar.test.tsx
```diff
@@ -1,8 +1,17 @@
...
 
 import LeftSidebar from '@/layout/LeftSidebar';
 
+jest.mock('next/link', () => jest.fn(
+  (
+    props: LinkProps & { children?: React.ReactNode } & React.HTMLAttributes<HTMLAnchorElement>,
+  ) => (
+    <a className={props.className} href={props.href.toString()}>{props.children}</a>
+  ),
+));
```
argaen, line 13: unrelated but deletes a warning of missing `act` due to async update of `ForwardRef(LinkComponent)`.

src/components/TransactionsTable.tsx
```diff
@@ -43,8 +43,12 @@ export default function TransactionsTable({
         },
         order: {
           fk_transaction: {
-            date: 'ASC',
+            date: 'DESC',
           },
+          // This is so debit is always before credit
+          // so we avoid negative amounts when display
+          // partial totals
```
argaen, line 50: this fixes all accounts but INCOME I think but I haven't seen a case yet. Will fix when it appers.

src/components/TransactionsTable.tsx
```diff
@@ -130,8 +134,8 @@ const columns: ColumnDef<Split>[] = [
       const currentRow = rows.findIndex(
         (otherRow: Row<Split>) => otherRow.original.guid === row.original.guid,
       );
-      const previousRows = rows.slice(0, currentRow + 1);
-      const totalPreviousSplits = previousRows.reduce(
+      const nextRows = rows.slice(currentRow, rows.length + 1);
+      const totalPreviousSplits = nextRows.reverse().reduce(
```
argaen, line 138: order is inverted after the changes above so need to reverse it to accumulate

PR #64, opened 2023-07-19
Fix double login

argaen:
Due to bad SWR management we were being prompted twice to login when the token expired. This fixes it by moving the error management within the fetcher function and disabling automatic re-validation. We now refresh every 100 seconds instead


src/hooks/useUser.ts
```diff
@@ -19,34 +19,45 @@ const emptyUser = {
...
+    isGapiLoaded ? '/api/user' : null,
+    getUser,
+    {
+      refreshInterval: 100000,
+    },
+  );
 
   React.useEffect(() => {
-    if (error) {
+    if (!isLoading && user && !user.isLoggedIn) {
```
argaen, line 52: we never have an error now as if there's an error in the fetcher we will return an empty user.

The isLoading part is so we only re-redirect once we have a result.

PR #65, opened 2023-07-19
Compress data file

argaen:
Saving the file to Google drive was taking ~20 seconds with a file size of close to 5MB. This is an improvement so we save data in a gzip compressed format using pako. This reduced the size to 1.8MB.


src/book/datasource.ts
```diff
@@ -41,6 +42,7 @@ export async function initDB(database: Uint8Array): Promise<DataSource> {
   return datasource;
 }
 
+// TODO: Need to mutate the accounts SWR key so Root is shown!
```
argaen, line 45: Unrelated but something that came to my mind while working on this.

PR #70, opened 2023-07-21
Saving indicator and save after modification

### LINKED ISSUE #26: Improving saving strategy
Currently we have a `Save` button that users have to click to make sure the data is uploaded to Google Drive. The problem though is that it's risky as if the user closes the tab all data is lost. We should periodically save the data (i.e. every minute) and also ask the user before leaving the page if they want to save the data before closing so they don't loose data.

There are some things to take into account:

- Saving takes few seconds, depends on how big the user file is so we will need to make sure the schedule is bigger than that.
- When asking the user to save, we should display some animation that waits for the file to be saved and then shows success or error.

argaen:
This PR adds some UI changes so users see when saving is happening. 

Screenshot 2023-07-21 at 5 39 00 PM

Not only this but I've implemented some state management to know when saving is happening as now it can happen from different points in the code. (i.e. adding a new account, importing a new book, etc). The state management is implemented using SWR. I know it may feel a bit hacky but I've tried the following approaches without success:

- Using custom hooks doesnt work because the result if the hook is local to the component rendering it. This means the state change only happens when you push the Save button but when doing it in other UI components, the SaveButton component doesn't realize about that.
- Moving the datasource to a context. This one worked but the problem is that the datasource is used in multiple parts of the code and thus, whenever the flag for "isSaving" changed, it re-rendered components that didn't need to be re-rendered (i.e. accounts table and so on which only need the `isLoaded` flag).

With this new state management we can be granular and only change and re-render specifically what we need.

In order to not mix state management and data fetching, I'm splitting the SWR keys as `/api/<whatever` and `/state/<whatever>`.

As mentioned in #26, we may find a point where saving on each change is a bit costly due to latency issues. We can accumulate changes locally and save periodically in the future but for now it works so leaving as is.

Closes #26

### DIFF
--- a/src/app/dashboard/accounts/[guid]/page.tsx
+++ b/src/app/dashboard/accounts/[guid]/page.tsx
@@ -9,6 +9,7 @@ import { Split } from '@/book/entities';
 import TransactionsTable from '@/components/TransactionsTable';
 import AddTransactionButton from '@/components/buttons/AddTransactionButton';
 import { getAccountsWithPath } from '@/book/queries';
+import { useDataSource } from '@/hooks';
 
 export type AccountPageProps = {
   params: {
@@ -17,6 +18,7 @@ export type AccountPageProps = {
 };
 
 export default function AccountPage({ params }: AccountPageProps): JSX.Element {
+  const { save } = useDataSource();
   let { data: accounts } = useSWRImmutable(
     '/api/accounts',
     getAccountsWithPath,
@@ -90,7 +92,10 @@ export default function AccountPage({ params }: AccountPageProps): JSX.Element {
         <div className="col-span-2 col-end-13 justify-self-end">
           <AddTransactionButton
             account={account}
-            onSave={() => mutate(`/api/splits/${params.guid}`)}
+            onSave={() => {
+              save();
+              mutate(`/api/splits/${params.guid}`);
+            }}
           />
         </div>
       </div>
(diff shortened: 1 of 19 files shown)


src/app/dashboard/accounts/[guid]/page.tsx
```diff
@@ -90,7 +92,10 @@ export default function AccountPage({ params }: AccountPageProps): JSX.Element {
         <div className="col-span-2 col-end-13 justify-self-end">
           <AddTransactionButton
             account={account}
-            onSave={() => mutate(`/api/splits/${params.guid}`)}
+            onSave={() => {
+              save();
```
Comment (argaen) [inline src/app/dashboard/accounts/[guid]/page.tsx:96]: simple! it basically exports the DB and uses the storage instance to upload the data.

src/components/AccountsTable.tsx
```diff
@@ -31,10 +31,7 @@ export default function AccountsTable(
 
   const root = accounts.find(a => a.type === 'ROOT');
   if (root && !todayPrices.isEmpty) {
-    const start = performance.now();
     rows = buildNestedRows(root, accounts, todayPrices).subRows;
-    const end = performance.now();
-    console.log(`build nested rows: ${end - start}ms`);
```
argaen, line 37: The cost is 40ms average so removing this as it's not super critical compared to others

src/components/buttons/ImportButton.tsx
```diff
@@ -39,58 +26,25 @@ export default function ImportButton(): JSX.Element {
...
         const rawBook = new Uint8Array(loadEvent.target.result as ArrayBuffer);
-        await saveBook(rawBook, storage, datasource);
+        await importBook(rawBook);
       }
     };
 
     fileReader.readAsArrayBuffer(event.target.files[0]);
   }
 }
-
```
argaen, line 69: This logic is now part of the useDataSource hook.

PR #102, opened 2023-07-28
Fix splits balance validator

### LINKED ISSUE #101: Split difference validator not correct
Seems there may be some decimal problem:

Screenshot 2023-07-28 at 10 50 31 AM

argaen:
Fixes #101 

As an extra, it also fixes another bug where we would call `getExchangeRate` for values of 0 which would trigger errors and disallow us to add dividend splits

### DIFF
--- a/src/__tests__/components/forms/transaction/__snapshots__/SplitField.test.tsx.snap
+++ b/src/__tests__/components/forms/transaction/__snapshots__/SplitField.test.tsx.snap
@@ -125,7 +125,6 @@ exports[`SplitField renders with empty data 1`] = `
         class="w-full pl-4 text-right bg-gunmetal-800 border-none rounded-sm m-0"
         disabled=""
         name="splits.0.quantity"
-        placeholder="0"
         step="0.001"
         type="number"
       />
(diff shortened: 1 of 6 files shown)


src/__tests__/components/forms/transaction/__snapshots__/SplitField.test.tsx.snap
```diff
@@ -125,7 +125,6 @@ exports[`SplitField renders with empty data 1`] = `
         class="w-full pl-4 text-right bg-gunmetal-800 border-none rounded-sm m-0"
         disabled=""
         name="splits.0.quantity"
-        placeholder="0"
```
argaen, line 128: removing as it adds confusion not knowing if it's populated with a 0 or just a placeholder

src/components/forms/transaction/SplitField.tsx
```diff
@@ -90,7 +90,7 @@ export default function SplitField({
         <span
           className="absolute px-1 py-2"
         >
-          {currencyToSymbol(account?.commodity?.mnemonic || '')}
+          {currencyToSymbol(account?.commodity?.mnemonic || '').slice(0, 3)}
```
argaen, line 93: fixes layout for when IDs are too long. I.e.:

Screenshot 2023-07-28 at 11 18 28 AM

PR #104, opened 2023-07-31
Support for dividends in different currencies

argaen:
In a change we introduced in https://github.com/maffin-io/maffin-app/commit/dca62c03ab2f67abd681772219b1a9f0a81a4f07#diff-28a1eddf39fdf2988e181547dae13df9391fab5a9623fa0df533fc11ef8ea214R75 we changed the way the transaction's currency is selected. Now, if one of the accounts has mainCurrency as currency, that's the transaction's currency. This created some issues with the previous code we had for storing dividends and this solves it.

### DIFF
--- a/src/book/models/InvestmentAccount.ts
+++ b/src/book/models/InvestmentAccount.ts
@@ -280,18 +280,30 @@ export default class InvestmentAccount {
    */
   _dividend(split: Split): void {
     split.transaction.splits.forEach(otherSplit => {
-      if (split.guid !== otherSplit.guid && otherSplit.quantity > 0) {
+      if (
+        split.guid !== otherSplit.guid
+        && this.currency === otherSplit.account.commodity.mnemonic
+        && otherSplit.quantity > 0
+      ) {
         const value = new Money(
-          otherSplit.value,
-          split.transaction.currency.mnemonic,
+          otherSplit.quantity,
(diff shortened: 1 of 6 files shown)


src/book/models/InvestmentAccount.ts
```diff
@@ -280,18 +280,30 @@ export default class InvestmentAccount {
    */
   _dividend(split: Split): void {
     split.transaction.splits.forEach(otherSplit => {
-      if (split.guid !== otherSplit.guid && otherSplit.quantity > 0) {
+      if (
+        split.guid !== otherSplit.guid
+        && this.currency === otherSplit.account.commodity.mnemonic
+        && otherSplit.quantity > 0
```
argaen, line 286: instead of selecting a random account with different guid, we pick the account that matches the currency of this account

src/components/buttons/TransactionFormButton.tsx
```diff
@@ -44,7 +44,6 @@ export default function TransactionFormButton(
           action={action}
           onSave={() => {
             save();
-            (defaultValues?.splits || []).forEach(split => mutate(`/api/splits/${split.account.guid}`));
```
argaen, line 47: unrelated to this change, moved the logic to TransactionForm.

PR #108, opened 2023-08-01
Home page revamp

argaen:
Initial revamp of the home page. Changes done include:

- Added a date range picker so the splits being parsed are filtered.
- Changed the AccountsTable to be visually better, include expanding/collapsing children, etc.

Others:

- I've fixed a login bug where `/api/user` key was not being revalidated even after mutating
- minor changes to `AddAccountButton` component to make it consistent with the transaction one.


src/app/dashboard/accounts/[guid]/page.tsx
```diff
@@ -122,13 +122,9 @@ export default function AccountPage({ params }: AccountPageProps): JSX.Element {
   });
 
   const split1 = new Split();
-  split1.value = 0;
-  split1.quantity = 0;
   split1.fk_account = account;
 
   const split2 = new Split();
-  split2.value = 0;
-  split2.quantity = 0;
```
Comment (argaen) [inline src/app/dashboard/accounts/[guid]/page.tsx:131]: This fixes the "bug" in the UI where we had to type twice for entering a quantity/value!

src/app/dashboard/accounts/page.tsx
```diff
@@ -1,21 +1,44 @@
...
+      earliestDate || DateTime.now().startOf('year'),
+      DateTime.now(),
+    ),
+  );
+
   let { data: accounts } = useSWRImmutable(
     '/api/accounts/splits',
     () => getAccountsWithPath({
-      relations: { splits: true },
+      relations: { splits: { fk_transaction: true } },
```
argaen, line 41: The data is needed so we can filter transactions/splits when the date ranges change.

src/app/dashboard/accounts/page.tsx
```diff
@@ -28,26 +51,94 @@ export default function AccountsPage(): JSX.Element {
...
 
+  let tree: AccountsTree = {
+    account: {},
+    total: new Money(0, 'EUR'),
+    children: [],
+  };
+  const root = accounts.find(a => a.type === 'ROOT');
+  if (root && !todayPrices.isEmpty) {
+    tree = buildNestedRows(root, accounts, todayPrices, dateRange);
+  }
```
argaen, line 62: We build the tree in the page rather than in the AccountsTable because the tree will be needed for rendering other components in the future version of the page.

src/app/dashboard/accounts/page.tsx
```diff
@@ -28,26 +51,94 @@ export default function AccountsPage(): JSX.Element {
...
+  current: Account,
+  accounts: Account[],
+  todayQuotes: PriceDBMap,
+  dateRange: Interval,
+): AccountsTree {
+  const { childrenIds } = current;
+  let total = current.getTotal(dateRange);
+  let subRows: AccountsTree[] = [];
+
+  if (current.type === 'ROOT') {
```
argaen, line 107: Hmmm to try in the future but we could actually compute the total net worth in the Root account. Use mainCurrency as currency.

src/components/DateRangeInput.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import Datepicker from 'react-tailwindcss-datepicker';
```
argaen, line 2: new dependency. Sadly it's not as customisable as I would like but for now it offers the functionality we need, we can always contribute in the future.

src/components/pages/account/TotalLineChart.tsx
```diff
@@ -29,14 +29,12 @@ export default function TotalLineChart({
   });
 
   return (
-    <div className="bg-gunmetal-700 rounded-sm my-6 mr-6">
```
argaen, line 32: The "card" div is more a layout concern so moving it outside.

src/hooks/useUser.ts
```diff
@@ -45,6 +45,7 @@ export default function useUser(): { user: User } {
     getUser,
     {
       refreshInterval: 100000,
+      revalidateOnMount: true,
```
argaen, line 48: this fixes the bug of not re-validating after we log-in even though we called `mutate`

PR #111, opened 2023-08-02
Hide SWR behind explicit API

argaen:
The approach for hiding async functions that components need behind SWR is becoming the default way of accessing data in this project. One of the things that was worrying me was that the keys are arbitrary and can be added as many we want but without realising if there is shared functionality between them or not. With this change we add a layer on top of SWR so keys are standard and gives an easy way to identify already existing functionality.

With this, we get a central repo of functions used by components that have common caching so make sure we re-query as less times as possible.

In future version I may add a version on top of global `mutate` just to make sure we always point to keys that indeed exist.


src/app/dashboard/accounts/[guid]/page.tsx
```diff
@@ -22,43 +23,8 @@ export type AccountPageProps = {
...
-          quantityNum: 'ASC',
-        },
-      });
-      const end = performance.now();
-      console.log(`get splits: ${end - start}ms`);
-      return sps;
-    },
-  );
+  let { data: accounts } = useApi('/api/accounts') as SWRResponse<Account[]>;
+  let { data: splits } = useApi('/api/splits/<guid>', { guid: params.guid }) as SWRResponse<Split[]>;
```
Comment (argaen) [inline src/app/dashboard/accounts/[guid]/page.tsx:27]: way simpler!

src/hooks/useApi.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Wrapper around SWR to make the usable keys explicit and re-usable
+ */
+export default function useApi(k: ApiPaths | null, args?: any): SWRResponse {
+  let f = k !== null ? API[k].f : () => {};
+  const swrArgs = k !== null ? API[k].swrArgs : undefined;
+
+  let key: string | null = k;
+  if (k === '/api/splits/<guid>') {
+    key = k.replace('<guid>', args.guid);
+    f = () => queries.getSplits(args.guid);
```
argaen, line 19: This is quite ugly, at some point would be good to investigate how to make a proper router with proper types...

Issue #84, opened 2023-07-25
Accounts page revamp


argaen:
Currently we are only showing the accounts tree. We should display some useful information like:

- accounts tree (collapse/etc and navigate to specific page).
- income/expenses graph
- net worth graph
- Assets radar with percentages of which assets contribute to it.
- small widget with last transactions

argaen:
Accounts tree was added in https://github.com/maffin-io/maffin-app/pull/108

argaen:
Added net worth in #112

argaen:
A simple one to add is "latest movements" to have a quick view of latest transactions happening, maybe show the latest 5 or something like that.

argaen:
After all the changes above, home page currently includes the following:

- account tree table navigation with totals
- net worth donut chart
- net worth bar chart
- latest movements
- expenses histogram
- income histogram

Closing this for now as it feels quite complete

PR #117, opened 2023-08-04
Add net worth histogram

argaen:
The PR adds a net worth histogram calculated from monthly income - expenses, accumulating net profit. 

Screenshot 2023-08-04 at 6 33 36 PM

The calculation uses a different approach from the piechart net worth which uses assets - liabilities. With my numbers, I'm getting a difference of 3k and I don't know why, maybe I'm missing something or maybe some accounting is done wrong somewhere... will investigate with simpler books to see if this is repeated.

Relates to #84

### DIFF
--- a/src/components/pages/accounts/NetWorthHistogram.tsx
+++ b/src/components/pages/accounts/NetWorthHistogram.tsx
@@ -1,0 +1,217 @@
+import React from 'react';
+import { DateTime, Interval } from 'luxon';
+
+import Chart from '@/components/charts/Chart';
+import type { AccountsTree } from '@/types/accounts';
+import { moneyToString } from '@/helpers/number';
+
+export type NetWorthHistogramProps = {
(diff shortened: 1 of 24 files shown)


src/components/pages/accounts/NetWorthHistogram.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+export default function NetWorthHistogram({
+  startDate,
+  tree,
+}: NetWorthHistogramProps): JSX.Element {
+  const interval = Interval.fromDateTimes(
+    startDate || DateTime.now(),
+    DateTime.now(),
+  );
+  const dates = interval.splitBy({ month: 1 }).map((d) => (d.start as DateTime).plus({ month: 1 }));
```
argaen, line 21: We should replicate this logic to all the other charts rather than hardcoding months and use time series... would make the series more manageable.

PR #121, opened 2023-08-07
Add expenses/income histograms to home page

argaen:
Adds expenses/income histograms in the home page to see monthly contribution to those types. The histograms have one serie per children account in that type, only first level taken into account. In the future we can allow users to dig deeper into categories (i.e. "Leisure" one may contain children accounts).

The histogram reacts to the date selection too.

Relates to #84


src/components/pages/accounts/NetWorthHistogram.tsx
```diff
@@ -67,7 +67,7 @@ export default function NetWorthHistogram({
     incomeAccount = tree.children.find(a => a.account.type === 'INCOME');
     const expensesAccount = tree.children.find(a => a.account.type === 'EXPENSE');
     dates.forEach(date => {
-      const monthYear = (date as DateTime).minus({ month: 1 }).toFormat('MMM/yy');
```
argaen, line 70: I think this introduced a bug where net worth is not right in the histogram

PR #132, opened 2023-08-08
Data api refactor

argaen:
I wanted to simplify a bit the way we use the typeorm entities as I felt like I was starting to add to many permutations of loading the same data (`/api/accounts`, `/api/accounts/guid`, etc). I've simplified and now we have `/api/accounts` which loads the accounts with their splits. This means currently we've slowed down the initial start up time but reduced the time for other pages + reduced code complexity.

Some learnings while working on this:

- Adding relationId fields to entities has its performance cost.
- Using `select` in the find typeorm operations is less costly than loading the whole relationship (obviously).
- Seems loading the entities separately and merging them later may be less costly than loading entities and their relations. I don't have facts for this as I was only measuring the query and entity creation time but not the added time later for linking the entities together.

I'm thinking that in later versions I may explore an approach where the loading of each entity is separated (`/api/accounts`, `/api/splits`, etc) but then the API keeps some global objects and links them. We can also add endpoints like `/api/splits/<guid>` that access the global object.


argaen:
Closing in favor of #138

PR #137, opened 2023-08-10
Set accounts paths at import

argaen:
The account path is something we use a lot in our app. Until now we were using a recursive query that calculates the paths and we did that, because Gnucash schema (which we are based on) doesn't have a field to store that. This gives some problems as it's something we almost always want/need.

In this PR we are calculating the paths when the user imports a file and also when new accounts are saved/created. Note that with this, we are creating a new field in the Account schema.


src/hooks/useApi.ts
```diff
@@ -51,13 +51,7 @@ const API: {
...
   },
   '/api/accounts': {
-    f: queries.getAccountsWithPath,
-  },
-  '/api/accounts/splits': {
-    f: () => queries.getAccountsWithPath({
-      relations: { splits: { fk_transaction: true } },
-      showRoot: true,
-    }),
+    f: q.getAccounts,
```
argaen, line 54: yes, I'm deleting the extra `/api/accounts/splits`. This is related to the work in #132 as I'm trying to simplify the amount of operations/api entries we have. The use case is that we load the accounts at start both for showing data and the selector. Before this caused two queries that return mostly the same data. We just do one now.

PR #140, opened 2023-08-11
Use SQL aggregations

argaen:
So.. apparently loading entities into JS classes with typeorm is quite costly. In the beginning we were loading the home page by loading ALL splits and then aggregating their totals and monthly totals with code. Loading all splits with the needed joins was quite costly, in the order of seconds. After this PR, all the data we load is more manageable and the most expensive queries are in the order of 300-400ms in my machine. Some of the optimisations done here are:

- Aggregate monthly totals in SQL, this is in `lib/queries/getMonthlyTotals.ts`.
- Remove the total aggregation, we now calculate this by accumulating the monthly totals.
- the `/api/accounts` endpoint just returns an accounts map. We now again have a `/api/splits/<guid>` to load specific splits for a single account with the needed relations.
- Changed the API approach. Now we have specific hook for each of the queries. This allows us to make complex hooks (i.e. monthly totals depending on accounts one).
- New api/queries to return monthly totals (which returns aggregated with children)


src/book/Money.ts
```diff
@@ -61,10 +62,17 @@ export default class Money {
...
-    );
+    try {
+      return djs.toDecimal(
+        djs.transformScale(this._raw, scale),
+        ({ value, currency }) => moneyToString(Number(value), currency.code),
+      );
+    } catch {
+      return djs.toDecimal(
+        djs.transformScale(this._raw, scale),
+        ({ value, currency }) => `${value} ${currencyToSymbol(currency.code)}`,
```
argaen, line 73: stock commodities are not valid currencies so we fallback

PR #147, opened 2023-08-14
Fix commodities

argaen:
So for some reason, having inline `Commodity.find` passed to the fetcher directly fails and we need to encapsulate it within an arrow function. Probably some `this` context or something like that as it fails with `Commodity` being `undefined`


src/__tests__/lib/queries/getAccounts.test.ts
```diff
@@ -59,7 +59,6 @@ describe('getAccounts', () => {
     const accounts = await getAccounts();
 
     expect(accounts.root.guid).toEqual('a');
-    expect(accounts.a.guid).toEqual('a');
```
argaen, line 62: Related to the change to stop showing two ROOT accounts in the selector (The `template` one shows)

PR #200, opened 2023-09-09
Fix split re-usage adding transactions

argaen:
By passing the splits all the way from from the page component, the uuid of the split was being reused which means the transactions were mixing splits and re-associating them to the latest transaction. This fixes it.

Also fixed a bug when importing where SWR wasn't being restarted (set all to undefined now).


src/app/dashboard/accounts/[guid]/page.tsx
```diff
@@ -94,11 +93,6 @@ export default function AccountPage({ params }: AccountPageProps): JSX.Element {
     return true;
   });
 
-  const split1 = new Split();
-  split1.fk_account = account;
-
-  const split2 = new Split();
```
Comment (argaen) [inline src/app/dashboard/accounts/[guid]/page.tsx:100]: the split was created in the first render and then reused for every transaction you add from this page which is bad bad.

src/components/buttons/TransactionFormButton.tsx
```diff
@@ -56,7 +63,18 @@ export default function TransactionFormButton(
...
         className={className}
         onClick={async () => {
-          if (action === 'update' || action === 'delete') {
+          if (action === 'add') {
+            const split1 = new Split();
+            if (account) {
+              split1.fk_account = account;
+            }
+
+            const split2 = new Split();
```
argaen, line 72: now we re-create the split on every click of the "Add Transaction" button

src/hooks/api/index.ts
```diff
@@ -63,14 +63,17 @@ export function useAccountsMonthlyTotals(): SWRResponse<queries.MonthlyTotals> {
     fetcher(PriceDB.getTodayQuotes, '/api/prices/today'),
   );
 
-  const key = '/api/accounts/monthly-totals';
+  const key = (accounts && todayPrices) ? [
+    '/api/accounts/monthly-totals',
+    // We need to recompute when accounts and prices change
+    Object.keys(accounts),
+    Object.keys(todayPrices.keys),
+  ] : null;
```
argaen, line 71: with previous approach, even when accounts and prices changed, the monthly-totals were not being recalculated. By having a different key every time we have new data we make sure the monthly totals are updated to the latest.

src/hooks/useDataSource.ts
```diff
@@ -124,7 +124,13 @@ async function importBook(storage: BookStorage | null, rawData: Uint8Array) {
...
 
   await DATASOURCE.sqljsManager.loadDatabase(rawBook);
-  mutate((key: string) => key.startsWith('/api'));
+  mutate((key: string) => {
+    let k = key;
+    if (Array.isArray(key)) {
+      [k] = key;
+    }
+    return k.startsWith('/api');
+  }, undefined);
```
argaen, line 133: set to `undefined` so instead of having stale data we have no data. This is good because otherwise it causes having inconsistent accounts vs prices

PR #202, opened 2023-09-11
Fix transaction form split update

argaen:
With previous implementation, when you removed a split from an already existing transaction, that split was left dangling with still the original account associated but without a transaction which caused errors (when loading splits for that specific account for example).

This PR fixes that by deleting all splits with null transaction after an update. I considered an alternative implementation which was to delete the specific Split in SplitsField component, when the user clicks X but this means that the split is deleted before the user clicks "Update". If they don't update, the split is removed regardless which is bad.

### DIFF
--- a/src/__tests__/helpers/number.test.ts
+++ b/src/__tests__/helpers/number.test.ts
@@ -52,4 +52,8 @@ describe('moneyToString', () => {
   it('uses EUR by default if empty currency passed', () => {
     expect(moneyToString(1234567.89, '')).toEqual('€1,234,567.89');
   });
+
+  it('returns as expected when not a currency', () => {
+    expect(moneyToString(1234567.89, 'NVDA')).toEqual('1,234,567.89 NVDA');
+  });
 });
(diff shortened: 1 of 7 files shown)


src/__tests__/helpers/number.test.ts
```diff
@@ -52,4 +52,8 @@ describe('moneyToString', () => {
   it('uses EUR by default if empty currency passed', () => {
     expect(moneyToString(1234567.89, '')).toEqual('€1,234,567.89');
   });
+
+  it('returns as expected when not a currency', () => {
+    expect(moneyToString(1234567.89, 'NVDA')).toEqual('1,234,567.89 NVDA');
```
argaen, line 57: unrelated but fixes some exceptions in the account detail page where some graphs were failing to load due to some commodities not considered a valid currency

src/components/forms/transaction/TransactionForm.tsx
```diff
@@ -113,11 +114,14 @@ async function onSubmit(data: FormValues, action: 'add' | 'update' | 'delete', o
...
-  if (action === 'add' || action === 'update') {
+  if (action === 'add') {
     await transaction.save();
-  }
-
-  if (action === 'delete') {
+  } else if (action === 'update') {
+    await transaction.save();
+    await Split.delete({
+      fk_transaction: IsNull(),
```
argaen, line 122: I guess we could be more specific and search for the split guid difference but.. I can't think of an edge case where this is bad and as an extra we get a sanity check (we don't want splits with null transaction).

PR #213, opened 2023-09-13
Fix different currency dividends

argaen:
With the current logic, the case where an Account with commodity X and giving dividends in Y didn't work, dividends were being ignored because it didn't find the corresponding split to extract quantities/value from. This fixes it.

This PR adds some constraints that didn't exist before:

- The income account where you receive the dividends **must be** in main currency.
- You can't have an investment that started giving dividends in currency X and then changed to currency Y. This may have problems when people change brokers (some brokers do the conversion to your broker currency automatically). We can work on this once the problem appears.


src/components/pages/account/SplitsHistogram.tsx
```diff
@@ -26,14 +26,10 @@ export default function SplitsHistogram({
 
   if (splits.length) {
     const interval = Interval.fromDateTimes(
-      splits[splits.length - 1].transaction.date,
-      splits[0]?.transaction.date,
+      splits[splits.length - 1].transaction.date.startOf('year'),
+      DateTime.now().endOf('year'),
```
argaen, line 30: unrelated, fixes a bug where some accounts having data only in one year failed to load the histogram.

src/hooks/api/index.ts
```diff
@@ -80,10 +80,15 @@ export function useAccountsMonthlyTotals(): SWRResponse<queries.MonthlyTotals> {
...
-  return useSWRImmutable(
+  const result = useSWRImmutable(
     key,
     fetcher(queries.getInvestments, key),
   );
+  if (result.error) {
+    throw new Error(result.error);
+  }
+
+  return result;
```
argaen, line 91: unrelated, propagating errors so we realize

PR #222, opened 2023-09-19
Minor refactors

argaen:
Some refactors, the biggest one is making pages components to be self working (data is pulled from use hooks rather than passed from the parent page). This makes the page components cleaner and it is clearer what each component needs (plus in the future it let's us load components 1 by 1 once their data is ready)


src/components/forms/transaction/SplitsField.tsx
```diff
@@ -28,7 +28,8 @@ export default function SplitsField({
   React.useEffect(() => {
     if (
       splits.length === 2
-      && splits[0].account.commodity.guid === splits[1].account?.commodity?.guid
+      && splits[0].account
+      && splits[0].account?.commodity?.guid === splits[1].account?.commodity?.guid
```
argaen, line 32: this is so we can load transaction form without any split default

src/components/forms/transaction/TransactionForm.tsx
```diff
@@ -157,6 +157,7 @@ async function onSubmit(data: FormValues, action: 'add' | 'update' | 'delete', o
     mutate(`/api/splits/${split.account.guid}`);
   });
 
+  mutate('/api/monthly-totals');
```
argaen, line 160: new transactions affect monthly totals. Probably we can optimise by updating only the affected accounts but it's not easy because we have to propagate this total through all the parents. Maybe calling `getMonthlyTotals` query with only the accounts of a specific type would help a bit already.

src/lib/queries/getMonthlyTotals.ts
```diff
@@ -48,13 +48,17 @@ export default async function getMonthlyTotals(
...
+  accounts.root?.childrenIds.forEach(
+    (childId: string) => {
+      aggregateChildrenTotals(
+        accounts[childId],
+        accounts,
+        todayQuotes,
+        monthlyTotals,
+      );
+
+      monthlyTotals[accounts[childId].type.toLowerCase()] = monthlyTotals[childId];
```
argaen, line 60: same as in `getAccounts`, now we can access the first level asset, expense accounts with `monthlyTotals.asset`, `monthlyTotals.expense`, etc.

PR #229, opened 2023-09-20
Improve key management for monthly totals

argaen:
Make sure we refresh monthly-totals after adding a new transaction. This is necessary for future onboarding and also because now users can add new transactions from the home page which means we need to refresh home page graphs when this happens.


src/hooks/api/index.ts
```diff
@@ -63,16 +63,17 @@ export function useAccountsMonthlyTotals(): SWRResponse<queries.MonthlyTotals> {
...
-    Object.keys(accounts),
-    Object.keys(todayPrices.keys),
-  ] : null;
+  // This makes sure that every time new accounts appear we recalculate
+  // the monthly totals.
+  //
+  // Not only this but re-using the same key has super poor performance
+  // for some reason
+  const key = (accounts && todayPrices)
+    ? `/api/monthly-totals?ids=${Object.keys(accounts).join(',')}` : null;
```
argaen, line 72: not a big difference from before but prefer to have consistent type for keys (strings).

src/components/forms/transaction/TransactionForm.tsx
```diff
@@ -157,7 +157,7 @@ async function onSubmit(data: FormValues, action: 'add' | 'update' | 'delete', o
     mutate(`/api/splits/${split.account.guid}`);
   });
 
-  mutate('/api/monthly-totals');
+  mutate((key: string) => key.startsWith('/api/monthly-totals'));
```
argaen, line 160: this didn't work before because the key didn't exist as it was an array. This now properly refreshes the data.

PR #235, opened 2023-09-21
Option to add opening balance in bank account creation

argaen:
When creating a new bank account, there's an option now to add opening balance.


src/components/forms/account/AccountForm.tsx
```diff
@@ -125,7 +150,61 @@ export default function AccountForm({ onSave }: AccountFormProps): JSX.Element {
...
-  await Account.create({ ...data }).save();
-  mutate('/api/accounts');
+  const account = await Account.create({ ...data }).save();
+
+  mutate(
+    '/api/accounts',
+    async (accounts: AccountsMap) => {
+      const [child, parent] = await Promise.all([
+        Account.findOneByOrFail({ guid: account.guid }),
+        Account.findOneByOrFail({ guid: account.parent.guid }),
```
argaen, line 160: we load both accounts like this so childIds and parentId are loaded while parent is not loaded (otherwise we do memory overkill as parent contains an account with another parent, etc).

src/hooks/api/index.ts
```diff
@@ -63,17 +63,10 @@ export function useAccountsMonthlyTotals(): SWRResponse<queries.MonthlyTotals> {
     fetcher(PriceDB.getTodayQuotes, '/api/prices/today'),
   );
 
-  // This makes sure that every time new accounts appear we recalculate
-  // the monthly totals.
-  //
-  // Not only this but re-using the same key has super poor performance
-  // for some reason
```
argaen, line 70: so apparently the bad performance was due to apexcharts re-rendering multiple times on each key mutation. Now with chartjs is way better so we can use single key which is more manageable. When we make this query more granular (re-compute totals per account) we may re-add this.

PR #244, opened 2023-09-25
Remove apexcharts in favor of chartjs

argaen:
So apparently apexcharts is super slow when [updating](https://github.com/apexcharts/vue-apexcharts/issues/208) the data in a chart. The dashboard page was taking around 4-5s to re-render the page after adding a new transaction or account while consuming loads of CPU. I tried different approaches like moving the data fetching to the components directly so only the necessary ones are re-rendered but discovered that just re-rendering one single graph was still super slow.

After trying chart.js it's orders of magnitude faster so migrating to it. Also I'm quite happy with the library as the community and documentation is better.


grojguy:
Sadly, I had a similar experience. After a webpage with apexchart was open for several hours, it would become ridiculously slow to respond to any input... sometimes up to 10 seconds, or worse, crash the webpage. If I disabled the apexchart, and the problem was no more. I tried very hard to get apexcharts to work, but finally I had to abandon it. It was unusable for me, so I moved to chart.js also.  I thank the developer for their work on apexchart, but there is definitely a major latent issue.

argaen:
Yup! if it wouldn't have been for this problem I would've stayed with apexcharts, it worked. But this extra latency was crazy high and my app does a lot of updates to the data...

PR #270, opened 2023-10-09
Fix wrong tx currency for investment

argaen:
For transactions where main split was a STOCK the selected transaction currency was the STOCK commodity. This creates problems as investment accounts contain splits with different currency (sometimes it's the stock currency and others the STOCK commodity). This PR fixes that by selecting the commodity that is a CURRENCY from the transaction splits.


src/components/forms/transaction/SplitField.tsx
```diff
@@ -95,8 +95,10 @@ export default function SplitField({
                       split => split.fk_account.commodity.guid === mainCurrency?.guid,
                     )) {
                       form.setValue('fk_currency', mainCurrency);
-                    } else {
+                    } else if (splits[0].account.commodity.namespace === 'CURRENCY') {
                       form.setValue('fk_currency', splits[0].account.commodity);
+                    } else {
+                      form.setValue('fk_currency', splits[1].account.commodity);
```
argaen, line 101: this is the important part that fixes the bug

PR #199, opened 2023-09-09
New users onboarding

argaen:

PR #271, opened 2023-10-10
Onboarding v2

argaen:
New version for https://github.com/maffin-io/maffin-app/pull/199 after some of the changes have been introduced gradually in previous PRs.


src/components/forms/account/AccountForm.tsx
```diff
@@ -201,21 +207,7 @@ export default function AccountForm({ onSave }: AccountFormProps): JSX.Element {
...
-        Account.findOneByOrFail({ guid: account.parent.guid }),
-      ]);
-      return {
-        ...accounts,
-        [child.guid]: child,
-        [parent.guid]: parent,
-      };
-    },
-    { revalidate: false },
-  );
```
argaen, line 218: I added this code because I thought there were performance problems loading things from SWR but it was apexcharts re-rendering. I'm deleting this code now as it works fast with chart.js plus, it has an edge case for when the parent is a root placeholder (i.e. topmost assets, expenses, etc) where we should be updating `type_asset` too in here. It can be done but adds a bunch of complexity and this brings no gains right now

src/components/forms/commodity/CommodityForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+export type CommodityFormProps = {
+  onSave: Function,
+};
+
+const BASE_CURRENCIES = [
+  { label: 'EUR' },
+  { label: 'USD' },
+  { label: 'SGD' },
+  { label: 'GBP' },
+];
```
argaen, line 25: We probably should expand this as when it's single currency, it's just a name. Things get more complex when more currencies are added.

PR #284, opened 2023-10-12
Improve startup speed

argaen:
In previous versions, monthly-totals were calculated lazily which meant waiting for accounts and today prices to be ready (also lazily) and then wait for SWR to re-trigger. This meant few seconds of wait for the user to see the histograms/graphs. In this PR we add explicit loading of this data after building the datasource so the data can be see right after. There's a difference of few seconds as right now it's instant.


src/hooks/useDataSource.ts
```diff
@@ -64,6 +67,7 @@ export default function useDataSource(): DataSourceContextType {
           const end = performance.now();
           console.log(`init datasource: ${end - start}ms`);
         }
+        await preloadData();
```
argaen, line 70: preload data once we download the book

src/hooks/useDataSource.ts
```diff
@@ -106,7 +110,9 @@ async function importBook(storage: BookStorage | null, rawData: Uint8Array) {
   const rawBook = await importGnucashBook(parsedData);
   await DATASOURCE.sqljsManager.loadDatabase(rawBook);
 
+  // Remove all previous data from the state
   mutate((key: string) => key.startsWith('/api'), undefined);
+  await preloadData();
```
argaen, line 115: When importing the book, we also want to preload the data right after purging old data.

src/lib/queries/getMainCurrency.ts
```diff
@@ -1,32 +1,20 @@
...
- * we are providing this query that counts the commodity used for each INCOME/EXPENSE
- * account. The one most common is assigned as mainCurrency.
- *
- * This function assumes users will have their INCOME and EXPENSE usually with the same
- * currency. If for some reason this is not the case, if may be buggy and change
- * mainCurrency
- *
- * If we don't find one, returns null
+ * We consider the main currency to be the currency of the topmost
+ * ASSET account
```
argaen, line 5: Simplifying the main currency loading. We don't want it to change dynamically but just picking what the root asset account has. As future work, we should disable the users from being able to update placeholder accounts commodity (once we have an account update form).

PR #287, opened 2023-10-13
Add light theme

argaen:
This PRs adds behavior for detecting the system preference of the user set 'dark' or 'light' theme. If the user sets it manually, then it is stored in the localStorage for further sessions so it stays the same.

Note that we only store the theme when the user actively selects it. If a computer changes theme during the daytime (night mode), it will be respected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "react-tailwindcss-datepicker": "^1.6.6",
     "react-tooltip": "^5.21.5",
     "reflect-metadata": "^0.1.13",
+    "sharp": "^0.32.6",
     "simplebar-react": "^3.2.4",
     "sql.js": "^1.8.0",
     "swr": "^2.2.4",
--- a/src/components/ThemeButton.tsx
+++ b/src/components/ThemeButton.tsx
@@ -1,0 +1,55 @@
+import React from 'react';
+import { BiSolidMoon, BiSolidSun } from 'react-icons/bi';
+import { Tooltip } from 'react-tooltip';
+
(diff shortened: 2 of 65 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
     "react-tailwindcss-datepicker": "^1.6.6",
     "react-tooltip": "^5.21.5",
     "reflect-metadata": "^0.1.13",
+    "sharp": "^0.32.6",
```
argaen, line 38: Unrelated, re-adding coz removed trying to fix images and it was an issue with Next.js which after upgrading to newer version got fixed.

src/components/ThemeButton.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { BiSolidMoon, BiSolidSun } from 'react-icons/bi';
+import { Tooltip } from 'react-tooltip';
+
+export default function ThemeButton(): JSX.Element {
+  const preferredTheme = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+  const selectedTheme = localStorage.theme;
+  const [theme, setTheme] = React.useState<'dark' | 'light'>(selectedTheme || preferredTheme);
```
argaen, line 8: An improvement that could be done would be to set the theme in SWR like we do for /state/saving when saving the book. This would allow us to retrieve this state in ChartJS graph and change the gridlines and colors accordingly. Currently we have static values that may not work properly for light theme (or dark)

PR #293, opened 2023-10-14
Light theme fix

argaen:
Add support for Onboarding light theme.

Also fixed a bug that crashed loading data source when there was no data


src/hooks/useDataSource.ts
```diff
@@ -146,7 +146,7 @@ async function preloadData() {
     queries.getAccounts(),
     PriceDB.getTodayQuotes(),
   ]);
-  mutate('/api/main-currency', accounts.type_asset.commodity);
+  mutate('/api/main-currency', accounts.type_asset?.commodity);
```
argaen, line 149: when the book is empty, `type_asset` account does not exist.

PR #294, opened 2023-10-14
Fix form update

argaen:
Fix for #289

### DIFF
--- a/src/components/buttons/TransactionFormButton.tsx
+++ b/src/components/buttons/TransactionFormButton.tsx
@@ -95,6 +95,15 @@ export default function TransactionFormButton(
               ...tx,
               date: tx.date.toISODate() as string,
               fk_currency: tx.currency as Commodity,
+              // This is hacky but if we pass the Split
+              // class to the form, then we have reference errors as when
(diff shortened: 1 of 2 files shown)


src/components/buttons/TransactionFormButton.tsx
```diff
@@ -95,6 +95,15 @@ export default function TransactionFormButton(
...
               fk_currency: tx.currency as Commodity,
+              // This is hacky but if we pass the Split
+              // class to the form, then we have reference errors as when
+              // we update the form, it also updates the defaultValues
+              // which means formState.isDirty is not triggered properly
+              splits: tx.splits.map(split => ({
+                ...split,
+                value: split.value,
+                quantity: split.quantity,
+              } as Split)),
```
argaen, line 106: so apparently, when passing the Split object as defaultValues in useForm and useArrayFields, it's by reference. This means that when we update the form, the values are changed to the original object set in the defaultValues. This is problematic because react-hook-form calculates `formState.isDirty` by comparting `form.getValues()` with `defaultValues` which in this case it will be the same. This causes bugs when we update the form as the validation  and auto populate doesn't trigger

PR #295, opened 2023-10-14
Set latest split date when adding tx

### LINKED ISSUE #288: Autofill latest tx date for transaction form
it's quite annoying to have to change it every time back to the latest date you entered so we should auto populate with the latest TX date for the account

argaen:
When doing a run of adding expenses, you usually start from the oldest untracked and keep following chronologically. Currently, the date is auto populated to "now" but it's annoying to change that every time. This change makes that the latest transaction's in the account is populated in the form instead.

Closes #288

### DIFF
--- a/src/css/globals.css
+++ b/src/css/globals.css
@@ -117,7 +117,7 @@
   }
 
   .amount-negative {
-    @apply text-red-600 dark:text-red-500;
+    @apply text-red-600 dark:text-red-400/90;
   }
 
   .header {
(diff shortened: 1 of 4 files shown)


src/css/globals.css
```diff
@@ -117,7 +117,7 @@
   }
 
   .amount-negative {
-    @apply text-red-600 dark:text-red-500;
+    @apply text-red-600 dark:text-red-400/90;
```
argaen, line 120: unrelated

PR #302, opened 2023-10-16
Fix onboarding

argaen:
getAccounts was changed to include the root id in the accounts map. The problem with that is now we have 2 accounts when the book is initialized which broke the condition for when to show the onboarding

### DIFF
--- a/src/app/dashboard/accounts/page.tsx
+++ b/src/app/dashboard/accounts/page.tsx
@@ -34,7 +34,7 @@ export default function AccountsPage(): JSX.Element {
   }
 
   accounts = accounts || { root: { childrenIds: [] } };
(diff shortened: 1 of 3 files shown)


src/app/dashboard/accounts/page.tsx
```diff
@@ -34,7 +34,7 @@ export default function AccountsPage(): JSX.Element {
   }
 
   accounts = accounts || { root: { childrenIds: [] } };
-  const showOnboarding = Object.keys(accounts).length === 1;
+  const showOnboarding = Object.keys(accounts).length <= 2;
```
argaen, line 37: now we have `root` and `<rootGuid>`

PR #306, opened 2023-10-17
delete account feature

argaen:
Add support to delete accounts.

The button is only enabled if the account has no transactions


src/app/dashboard/accounts/[guid]/page.tsx
```diff
@@ -52,10 +51,9 @@ export default function AccountPage({ params }: AccountPageProps): JSX.Element {
 
   const account = accounts[params.guid] as Account;
   if (!account) {
-    router.push('/404');
```
Comment (argaen) [inline src/app/dashboard/accounts/[guid]/page.tsx:55]: This is.. kind of hacky. So apparently when deleting the account, this re-renders and then it triggers an error saying it can't change the component while rendering another one. This happened with deleting the account, mutating the accounts key and redirecting to `/dashboard/accounts`. After the error, it even goes to the 404 page.

This is a workaround to fix that... Will update once we find a better way.

PR #308, opened 2023-10-18
Fix instant refresh for onboarding opening balance

argaen:
I reordered the mutate calls and this made the monthly balance to update with a big delay. Re-ordering like this so it is instant.

### DIFF
--- a/src/components/forms/account/AccountForm.tsx
+++ b/src/components/forms/account/AccountForm.tsx
@@ -285,17 +285,19 @@ async function onSubmit(
 
   if (action === 'add') {
     await account.save();
+    mutate('/api/accounts');
     if (data.balance) {
       await createBalance(data, account);
     }
   } else if (action === 'update') {
     await account.save();
+    mutate('/api/accounts');
   } else if (action === 'delete') {
     await Account.remove(account);
+    mutate('/api/accounts');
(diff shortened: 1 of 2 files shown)


src/components/forms/account/AccountForm.tsx
```diff
@@ -352,4 +354,5 @@ async function createBalance(data: FormValues, account: Account) {
 
   // Opening balances affect net worth
   mutate('/api/monthly-totals', undefined);
+  mutate('/api/txs/latest', undefined);
```
argaen, line 357: latest transactions may be updated due to opening balance tx being one of the latest, so refreshing this.

src/components/forms/account/AccountForm.tsx
```diff
@@ -285,17 +285,19 @@ async function onSubmit(
...
   } else if (action === 'update') {
     await account.save();
+    mutate('/api/accounts');
   } else if (action === 'delete') {
     await Account.remove(account);
+    mutate('/api/accounts');
     router.replace('/dashboard/accounts');
   }
 
-  mutate('/api/accounts');
```
argaen, line 298: Refreshing accounts this late causes issues with `monthly-totals` taking very long to refresh. I'm... actually not sure why though

PR #309, opened 2023-10-18
Add commodity

argaen:
This PR changes the Commodity selector so it can create new commodities on the fly. The selector uses Stocker in the background to search for new symbols. If a symbol is not found, the commodity can still be created (for custom tracking) but there will be no automated price tracking.

There's a bug though now that we can create new commodities which is, we don't refresh `todayPrices` entry when a new commodity is added which causes an error when navigating back to home page. In a further PR we will fix this by refreshing it with the correct data.


amplify/backend/function/stockerlambda/ts/yahoo.ts
```diff
@@ -8,6 +8,51 @@ export class YahooError extends HTTPError {}
...
+    quotes.length === 0
+  ) {
+    throw new YahooError(
+      `No results for ${ticker}`,
+      404,
+      'NOT_FOUND',
+    );
+  }
+
+  const info = quotes[0];
```
argaen, line 45: We are returning only one but we could probably make it to return multiple ones... May change in the future

PR #320, opened 2023-10-19
Fix transactions table action buttons

argaen:
This was a nasty bug. So `TransactionFormButton` was being rendered with some defaultValues and then when re-rendered with new default values, they were not being applied because `React.useState` only applies the values on the first render. Changed the code so the passed default values are always used and applied

### DIFF
--- a/src/__tests__/components/buttons/TransactionFormButton.test.tsx
+++ b/src/__tests__/components/buttons/TransactionFormButton.test.tsx
@@ -94,7 +94,7 @@ describe('TransactionFormButton', () => {
     expect(TransactionForm).toHaveBeenLastCalledWith(
       {
         action,
-        defaultValues: {},
+        defaultValues: undefined,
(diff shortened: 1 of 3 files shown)


src/__tests__/components/buttons/TransactionFormButton.test.tsx
```diff
@@ -228,6 +229,53 @@ describe('TransactionFormButton', () => {
...
+      />,
+    );
+
+    const button = await screen.findByRole('button', { name: /add transaction/i });
+    fireEvent.click(button);
+
+    await waitFor(() => expect(TransactionForm).toHaveBeenLastCalledWith(
+      expect.objectContaining({
+        defaultValues: expect.objectContaining({
+          description: 'haha',
```
argaen, line 272: Before this change, the value was `hello` as `React.useState` was setting the state only on the first render and subsequent renders were not updating its value.

PR #326, opened 2023-10-20
Stocker refactor and staging

argaen:
Move stocker to use typescript and refactor a bit. Also while working on this yahoo quoteSummary endpoint stopped working so refactored to always use the `chart` endpoint instead.


.gitignore
```diff
@@ -1,5 +1,7 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
+amplify/backend/function/stockerlambda/src/*
```
argaen, line 3: This now contains the `build` result. The source code lives in `ts` folder

PR #337, opened 2023-10-24
Version upgrade

argaen:


src/components/buttons/AccountFormButton.tsx
```diff
@@ -31,6 +31,7 @@ export default function AccountFormButton({
 
   return (
     <>
+      {/* @ts-ignore */}
       <Modal
```
argaen, line 35: Wasn't able to fix the error that appears with the new version of types/react... will re-visit at some point once time passes

PR #376, opened 2023-11-15
Remove duplicate root

argaen:
Account selector was showing duplicate Root entry due to the accounts map containing the root account by guid and `root`. Changed `root` to `type_root` so it gets filtered out accordingly.

Also upgrade to packages latest versions


.nvmrc
```diff
@@ -1 +1 @@
-v16.17.0
+v18.17.0
```
argaen, line 1: version was inconsistent, fixing it.

src/lib/queries/getAccounts.ts
```diff
@@ -5,24 +5,23 @@ import type { AccountsMap } from '@/types/book';
...
+ * with type, same for the root account.
  */
 export default async function getAccounts(): Promise<AccountsMap> {
   const accounts = await Account.find();
 
   const accountsMap: AccountsMap = {};
   accounts.forEach(account => {
     if (account.type === 'ROOT' && !account.name.startsWith('Template')) {
-      accountsMap.root = account;
+      accountsMap.type_root = account;
```
argaen, line 17: The actual fix. Note this has impact to how the map is used, now to get the root account you need to access `type_root`

PR #398, opened 2023-11-17
Add migration class name

argaen:


src/book/migrations/1697688851711-AddCommodityName.ts
```diff
@@ -22,3 +22,5 @@ export class AddCommodityName1697688851711 implements MigrationInterface {
     'unimplemented';
   }
 }
+
+Object.defineProperty(AddCommodityName1697688851711, 'name', { value: 'AddCommodityName1697688851711' });
```
argaen, line 26: To fix `810-a78d418ce931fe02.js:70 Uncaught (in promise) en: e migration name is wrong. Migration class name should have a JavaScript timestamp appended.`

PR #420, opened 2023-11-28
Add Price button and form

argaen:
With this we can now add and edit prices. If you add a price with the same commodity -> currency and same date, it overrides the existing one.


src/app/dashboard/commodities/[guid]/page.tsx
```diff
@@ -56,6 +57,7 @@ export default function CommodityPage({ params }: CommodityPageProps): JSX.Eleme
       byCurrency[price.currency.guid].push(price);
     });
   }
+  const commodityCurrency = commodity.namespace !== 'CURRENCY' ? prices?.[0].fk_currency || undefined : undefined;
```
Comment (argaen) [inline src/app/dashboard/commodities/[guid]/page.tsx:60]: For non CURRENCY commodities, we always have the same currency assigned (i.e. STOCKS only have one currency). This takes the first currency for a stock and passes it as a default

PR #424, opened 2023-11-29
Form button

argaen:
No functional changes, this PR reduces duplication of having to create a button for each different form. We now have a generic button that can be used for any form (gets injected as children). This is better because now we have more control over the form in each component (pass different parameters).

PS: I've made the `getSplits` function slower on purpose to reduce complexity. In the future, we should add pagination to getSplits and some way to calculate total (to show the proper balance).


src/app/dashboard/accounts/[guid]/page.tsx
```diff
@@ -115,38 +116,56 @@ export default function AccountPage({ params }: AccountPageProps): JSX.Element {
...
+              )}
+            >
+              <TransactionForm
+                defaultValues={
+                  {
+                    date: (latestDate || DateTime.now()).toISODate() as string,
+                    description: '',
+                    splits: [Split.create({ fk_account: account }), new Split()],
+                    fk_currency: account.commodity,
+                  }
```
Comment (argaen) [inline src/app/dashboard/accounts/[guid]/page.tsx:136]: It's more verbose but we gain flexibility

PR #495, opened 2024-01-05
Add investment stats to account page

argaen:
Adds investment stats (like latest known price, avg price, etc.) for accounts that are investments


src/__tests__/components/pages/account/SplitsHistogram.test.tsx
```diff
@@ -20,6 +20,7 @@ jest.mock('@/hooks/api', () => ({
 describe('SplitsHistogram', () => {
   beforeEach(() => {
     jest.spyOn(apiHook, 'useSplits').mockReturnValue({ data: undefined } as SWRResponse);
+    jest.spyOn(DateTime, 'now').mockReturnValue(DateTime.fromISO('2023-12-31') as DateTime<true>);
```
argaen, line 23: change of year made tests fail, this fixes it

PR #505, opened 2024-01-07
Dont ask for custom commodity prices

argaen:


src/book/prices/PriceDB.ts
```diff
@@ -127,8 +127,11 @@ async function getCurrencyQuotes(): Promise<Price[]> {
 }
 
 async function getInvestmentQuotes(): Promise<Price[]> {
-  const commodities = await Commodity.findBy({
-    namespace: Not('CURRENCY'),
+  const commodities = await Commodity.find({
+    where: [
+      { namespace: 'STOCK' },
+      { namespace: 'FUND' },
+    ],
```
argaen, line 134: We used to assume that anything not 'CURRENCY' can be queried in yahoo finance but this is not the case. We will need to work in the future on this

src/app/dashboard/commodities/[guid]/page.tsx
```diff
@@ -58,7 +58,7 @@ export default function CommodityPage({ params }: CommodityPageProps): JSX.Eleme
       byCurrency[price.currency.guid].push(price);
     });
   }
-  const commodityCurrency = commodity.namespace !== 'CURRENCY' ? prices?.[0].fk_currency || undefined : undefined;
+  const commodityCurrency = commodity.namespace !== 'CURRENCY' ? prices?.[0]?.fk_currency || undefined : undefined;
```
Comment (argaen) [inline src/app/dashboard/commodities/[guid]/page.tsx:61]: when we add a new commodity there are no prices

PR #513, opened 2024-01-09
Improve api performance

argaen:
Change queries API so we can query single instances rather than having to query all of them and then accessing the specific one. This improves speed when loading detail pages of single instances.


src/hooks/api/index.ts
```diff
@@ -28,13 +28,34 @@ export function useMainCurrency(): SWRResponse<Commodity> {
...
+  if (result.error) {
+    throw new Error(result.error);
+  }
+
+  if (result.data && !guid) {
+    result.data.forEach(
+      (commodity: Commodity) => mutate(
+        `/api/commodities/${commodity.guid}`,
+        commodity,
+        { revalidate: false },
```
argaen, line 53: this is necessary or otherwise we get an infinite loop when rendering a detail page for some cases where we have a page asking for the list and the detail keys (i.e AccountSelector which is loaded in the Topbar)

PR #515, opened 2024-01-09
Support custom investments

argaen:
**Implementation**

Brings some big changes to how much we depend on Stocker, the API that retrieves prices from yahoo. In the past year it's been three times that the API has changed or was down or whatever. When this happened, maffin was not usable as we heavily depended on the data retrieved from there.

With the changes introduced here, we are always using offline data stored in `Price` entity and Stocker is only used on load time to retrieve today prices but if it fails, we will just use old data.

This decoupling allows us to track "custom" investments too. A custom investment is anything that has a non CURRENCY commodity and that doesn't exist in common investment APIs (i.e. yahoo). For example let's say you want to track how many Cows (yes the animal) you have, how price evolves, etc. Now you can do so by creating the COW commodity and then entering prices manually.

**Features**
- Works with yahoo being down
- Can add custom commodities and offline prices
- Better resolution when showing investment chart in the account detail for "young" investments
- Fix bug in Latest transactions showing wrong currency when the asset account is not in main currency
- Transaction form refactored, now it uses offline prices always so it doesn't contact stocker.
- Transaction form now supports having investment transactions where each split has a different currency, this is helpful for tracking fees.

**Things to improve**

Currently, in the investments page we still show only Account types that are STOCK or MUTUAL and same for account detail page, the COW account page would be shown as a generic instead of an investment one. I need to think a bit more on where to draw the line and how to call these new accounts... they are not STOCK/MUTUAL so maybe have to create a new type and differ from gnucash implementation...

**Futurology**

I see a future where Stocker doesn't exist and there are some plugins out there (or extensions) that allow you to track prices of random things you want, to do so, Maffin could expose an API so you can create new Price entities within the DB. Chrome extensions would be a way and would work directly with yahoo finance and others as we can avoid CORs errors when using Chrome extensions but then.. what about other browsers? Needs some more research, for now I'm happy with Maffin not crashing when Yahoo does.


src/apis/Stocker.ts
```diff
@@ -1,39 +1,85 @@
...
-    queryStringParameters: {
-      ids: [tickers].toString(),
-    },
-  };
-  const resp = await API.get(API_NAME, '/api/prices', options);
+/**
+ * Connect to Stocker API and retrieve current prices for
+ * all currency pairs and commodities.
+ */
+export async function insertTodayPrices(): Promise<void> {
```
argaen, line 18: This is very ugly but I want it this way, self-contained so we decouple Stocker as much as possible from the rest of the code. This goes in line of having in the future "extensions" or "plugins" that scan for prices

PR #525, opened 2024-01-11
Fix import book swr keys issue

argaen:
The specific account keys i.e. `/api/accounts/<guid>` do change when importing a new book because the accounts are different. With the previous version we were setting the keys to undefined but that keeps the key in the cache and may cause issues. With the current change we iterate over the cache and then delete the keys that belong to specific accounts and then we just mutate as undefined

### DIFF
--- a/src/hooks/useDataSource.ts
+++ b/src/hooks/useDataSource.ts
@@ -1,7 +1,7 @@
 import React from 'react';
 import { DataSource } from 'typeorm';
 import initSqlJs from 'sql.js';
-import { mutate } from 'swr';
+import { useSWRConfig, mutate, Cache } from 'swr';
 import pako from 'pako';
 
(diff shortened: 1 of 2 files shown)


src/hooks/useDataSource.ts
```diff
@@ -114,8 +115,12 @@ async function importBook(storage: BookStorage | null, rawData: Uint8Array) {
   const rawBook = await importGnucashBook(parsedData);
   await DATASOURCE.sqljsManager.loadDatabase(rawBook);
 
-  // Remove all previous data from the state
-  mutate((key: string) => key.startsWith('/api'), undefined);
```
argaen, line 118: This creates problems as we mutate keys that will not exist anymore and thus are `undefined` when SWR iterates over them

PR #555, opened 2024-01-25
Support dividend txs not in main currency

argaen:
After https://github.com/maffin-io/maffin-app/pull/515 we changed the tx currency for non currency related transactions to be the linked currency to that commodity. This meant that we can add txs in currency different than the main currency and that's something we didn't allow. This PR fixes the code in order to allow for those type of transactions.

### DIFF
--- a/src/book/models/InvestmentAccount.ts
+++ b/src/book/models/InvestmentAccount.ts
@@ -299,23 +299,36 @@ export default class InvestmentAccount {
    * doesnt support this by default. We rely on the user/maffin adding
    * an empty split that helps to identify the income/asset related
    * transaction.
+   *
+   * If the income split is in mainCurrency, we will retrieve the amountInCurrency
+   * from there. Otherwise, we use whatever priceDBMap.getPrice returns for that
+   * date.
    */
   _dividend(split: Split): void {
-    const originalSplit = split.transaction.splits.find(s => s.value === 0);
     const brokerSplit = split.transaction.splits.find(s => s.value > 0);
     if (!brokerSplit) {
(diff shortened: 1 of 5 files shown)


src/book/models/InvestmentAccount.ts
```diff
@@ -299,23 +299,36 @@ export default class InvestmentAccount {
...
+      s => s.account.type === 'INCOME' && s.account.commodity.mnemonic === this.mainCurrency,
+    );
+
+    const amountInCurrency = incomeSplit
+      ? incomeSplit.quantity * -1
+      : brokerSplit.quantity * this._priceDBMap.getPrice(
+        brokerSplit.account.commodity.mnemonic,
+        this.mainCurrency,
+        split.transaction.date,
+      ).value;
```
argaen, line 323: If there's an income split with the same currency as mainCurrency, then let's use the quantity from that split. If not, we transform the amount using PriceDBMap with whatever exchange rates are available

src/components/forms/transaction/MainSplit.tsx
```diff
@@ -45,6 +45,8 @@ export default function MainSplit({
           : prices.getPrice(account.commodity.mnemonic, txCurrency.mnemonic, d);
 
         setExchangeRate(rate);
+      } else {
+        setExchangeRate(Price.create({ valueNum: 1, valueDenom: 1 }));
```
argaen, line 49: Fixes a bug where when changing txCurrency back, it wouldn't update the exchangeRate back to 1 and thus setting wrong amounts in quantity/value field of the main split

PR #557, opened 2024-01-25
Change account types

argaen:
Related to https://github.com/maffin-io/maffin-app/discussions/556.

This PR adds:

- FIXED account type
- INVESTEMENT account type

and we delete STOCK and MUTUAL account type (this will cause with already existing accounts in this type but I've added a migration that if you re-import your file, it gets fixed automatically).


src/book/helpers/accountType.ts
```diff
@@ -1,10 +1,17 @@
 import type { Account } from '@/book/entities';
 
-export const ASSET_ACCOUNTS = ['ASSET', 'BANK', 'CASH', 'STOCK', 'MUTUAL', 'RECEIVABLE'];
+export const ASSET_ACCOUNTS = [
+  'ASSET',
+  'BANK',
+  'CASH',
+  'FIXED',
```
argaen, line 7: FIXED assets are assets that can't be sold easily like vehicles, land, buildings, etc.

src/book/helpers/accountType.ts
```diff
@@ -1,10 +1,17 @@
...
+  'INVESTMENT',
```
argaen, line 8: Anything a user would consider an investment and that it has non CURRENCY commodity

PR #558, opened 2024-01-25
Fix development server

argaen:
Upgrading to next 14.1.0 seems it breaks the development server. Rolling back


package.json
```diff
@@ -38,7 +38,7 @@
     "debounce-promise": "^3.1.2",
     "dinero.js": "^2.0.0-alpha.14",
     "luxon": "^3.4.4",
-    "next": "^14.1.0",
+    "next": "^14.0.4",
```
argaen, line 41: breaks with a message saying `useEffect is not exported for react`

PR #559, opened 2024-01-25
Demo site

argaen:
For hosting a demo site we need it to be isolated from users and from logins (as we will control this in the future). Thus it's the same exact page but with the following differences:

- No connection with Google oauth, we return a dummy user.
- The book is an example one downloaded from the same site hosted under `/public` folder. Tried to download from github but hitting CORs errors
- Latest prices from Stocker are not retrieved when initialising the data source (coz I don't know how much the cost of this would be).


amplify/backend/api/stocker/cli-inputs.json
```diff
@@ -1,26 +1,12 @@
...
       "permissions": {
         "setting": "open"
       }
-    },
-    "/api/search": {
-      "name": "/api/search",
-      "lambdaFunction": "stockerlambda",
-      "permissions": {
-        "setting": "open"
-      }
```
argaen, line 23: unrelated, just deleting old endpoints from stocker

package.json
```diff
@@ -38,7 +38,7 @@
     "debounce-promise": "^3.1.2",
     "dinero.js": "^2.0.0-alpha.14",
     "luxon": "^3.4.4",
-    "next": "^14.1.0",
+    "next": "^14.0.4",
```
argaen, line 41: downgrade as it breaks dev environment

src/app/user/login/page.tsx
```diff
@@ -40,7 +41,13 @@ export default function LoginPage(): JSX.Element {
     <button
       className="btn btn-primary"
       type="button"
-      onClick={() => { tokenClient.requestAccessToken(); }}
+      onClick={() => {
+        if (isDemo()) {
+          router.push('/dashboard/accounts');
+        } else {
+          tokenClient.requestAccessToken();
```
argaen, line 48: Don't show oauth if we are in demo

src/components/buttons/SaveButton.tsx
```diff
@@ -27,7 +28,7 @@ export default function SaveButton(): JSX.Element {
       id="save-button"
       type="button"
       className="btn btn-primary"
-      disabled={isSaving}
+      disabled={isSaving || isDemo()}
```
argaen, line 31: Make Save button not clickable as it would be confusing

src/hooks/useDataSource.ts
```diff
@@ -149,13 +150,15 @@ async function createEmptyBook() {
  */
 async function preloadData() {
   const accounts = await queries.getAccounts();
-  const mainCurrency = accounts.type_asset.commodity;
+  const mainCurrency = accounts.type_asset?.commodity;
```
argaen, line 153: Fixes bug with onboarding as we don't have this account yet

public/books/demo.sqlite.gz
argaen, line 1: Demo book file. This is going to be useful for future e2e tests where we deploy to demo/staging, run them to make sure site works and then deploy to production

.github/workflows/staging-deploy.yml
```diff
@@ -1,23 +1,26 @@
...
-          curl -X POST -d {} "${{ secrets.STAGING_WEBHOOK_URL }}" -H "Content-Type: application/json"
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - name: Opening pull request
+        id: pull
+        uses: tretuna/sync-branches@1.4.0
```
argaen, line 21: whenever we have a push to master, we will open a PR to `staging` branch that auto merges which will trigger a staging deploy. The order is not quite right but we can worry about this when we need to

PR #565, opened 2024-01-26
Fix bad syntax in build file

argaen:


.github/workflows/staging-deploy.yml
```diff
@@ -23,4 +23,4 @@ jobs:
           GITHUB_TOKEN: ${{secrets.GH_TOKEN}}
           FROM_BRANCH: "master"
           TO_BRANCH: "staging"
-          PULL_REQUEST_AUTO_MERGE_METHOD: "merge"
+          PULL_REQUEST_AUTO_MERGE_METHOD: "squash"
```
argaen, line 26: changes are not auto-merging, see if this fixes it

PR #599, opened 2024-02-06
Use auth0

argaen:
Move to using auth0 for managing users. Good because we can remove the `authorize` and `refresh` endpoints from Stocker and interact directly with auth0 using the `auth0-react` library.

Note we are doing some unofficial stuff to extract the Google access_token from Auth0 (https://github.com/maffin-io/maffin-blog/issues/259)


src/lib/auth0-provider.ts
```diff
@@ -0,0 +1,5 @@
+'use client';
+
+import { Auth0Provider } from '@auth0/auth0-react';
+
+export default Auth0Provider;
```
argaen, line 5: We need `use client` so the Auth0Provider works but we need to extract it so we can set metadata in the root page.

src/app/layout.tsx
```diff
@@ -37,7 +42,17 @@ export default function RootLayout({
     <html lang="en">
       <body>
         <noscript>You need to enable JavaScript to run this app.</noscript>
-        {children}
+        <Auth0Provider
+          domain={`maffin${isProd() ? '' : '-dev'}.eu.auth0.com`}
```
argaen, line 46: we have different tenants for development and production

src/app/layout.tsx
```diff
@@ -37,7 +42,17 @@ export default function RootLayout({
...
+          clientId={isProd() ? 'cEXnN96kEP3ER2EDJjmjRW0u2MEFBUKK' : 'mMmnR4NbQOnim9B8QZfe9wfFuaKb8rwW'}
```
argaen, line 47: each app in Auth0 has a different clientId

PR #603, opened 2024-02-06
Redirect to login if not auth

argaen:
Seems the Auth0Provider doesn't send to the login page when not authenticated so doing this manually in the dashboard layout

### DIFF
--- a/src/app/dashboard/layout.tsx
+++ b/src/app/dashboard/layout.tsx
@@ -9,14 +9,18 @@ import LeftSidebar from '@/layout/LeftSidebar';
 import Topbar from '@/layout/Topbar';
 import Loading from '@/components/Loading';
 import { useTheme } from '@/hooks/state';
+import useSession from '@/hooks/useSession';
+import { useRouter } from 'next/navigation';
 
 Modal.setAppElement('#modals');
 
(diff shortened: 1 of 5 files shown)


src/app/dashboard/layout.tsx
```diff
@@ -26,6 +30,12 @@ export default function DashboardLayout({
     }
   });
 
+  React.useLayoutEffect(() => {
+    if (!isLoading && !isAuthenticated) {
```
argaen, line 34: make sure we are not loading here or we will get a redirect to user/login even when we are auth in the first page load (because we still didn't retrieve the state)

src/hooks/useSession.tsx
```diff
@@ -12,14 +12,13 @@ const emptyUser: User = {
...
-  user: User;
-};
+} & Auth0ContextInterface<User>;
 
 /**
  * This hook captures Authorization data from Google Oauth2
  */
 export default function useSession(): SessionReturn {
-  const { user, isAuthenticated } = useAuth0();
+  const auth0 = useAuth0();
```
argaen, line 21: small refactor so `useSession` returns everything that `useAuth0` returns instead of having to expand manually

PR #610, opened 2024-02-07
Add pan functionality to home histograms

argaen:
Adding ability to pan using zoom chartjs plugin.

It's not the smoothest but best than nothing. In future version we probably want to add a secondary chart similar to stocks chart so you can scroll/select dates


public/books/demo.sqlite.gz
argaen, line 1: Re-uploading as it didn't update properly...

src/app/dashboard/accounts/page.tsx
```diff
@@ -19,6 +18,9 @@ import DateRangeInput from '@/components/DateRangeInput';
 import Onboarding from '@/components/onboarding/Onboarding';
 import * as API from '@/hooks/api';
 
+const NetWorthHistogram = dynamic(() => import('@/components/pages/accounts/NetWorthHistogram'), { ssr: false });
+const IncomeExpenseHistogram = dynamic(() => import('@/components/pages/accounts/IncomeExpenseHistogram'), { ssr: false });
```
argaen, line 22: need to import like this because zoom plugin needs to be frontend rendered

ruru-m07:
nice work!

ruru-m07:
Done!

PR #612, opened 2024-02-07
Update AccountForm.tsx

corinetiah:


argaen (changes requested):
Requesting changes!

also, can you update the test in https://github.com/maffin-io/maffin-app/blob/master/src/__tests__/components/forms/account/AccountForm.test.tsx#L306 by adding:

```
expect(mutate).toBeCalledWith('/api/investments')
```

to make sure we don't remove this line accidentally and we never realise

PR #614, opened 2024-02-08
Fix monthly total calculations

argaen:
The whole net worth and monthly totals was calculated the wrong way. We took a pragmatic approach of taking the latest exchange rates which would show the latest net worth (for current month) the right way. However, this means that for previous month, the net worth was wrong because the exchange rate of other commodities is probably different.

Current changes fix this by changing how we calculate monthlyTotals:

- For assets/liabilities, each month we aggregate the value of all previous months and then convert it using **that month exchange rate**
- For income/expense, we still use the same approach. Each month **only contains the same month totals** and we convert using that month exchange rate.

This is so we can answer this two questions:

- How much did I have in my different currency bank (in total) at this date?
- How much did I spent in this different currency account at this date?

Note the first question requires to accumulate all previous transactions to see the current value while the second only requires current month transactions

### DIFF
--- a/src/components/forms/price/PriceForm.tsx
+++ b/src/components/forms/price/PriceForm.tsx
@@ -95,6 +95,7 @@ export default function PriceForm({
                   placeholder="Convert to"
                   onChange={field.onChange}
                   namespace="CURRENCY"
+                  ignore={[(defaultValues?.fk_commodity as Commodity)?.guid || '']}
                   isDisabled={disabled || action !== 'add'}
                   defaultValue={defaultValues?.fk_currency as Commodity}
                 />
--- a/src/components/onboarding/Onboarding.tsx
+++ b/src/components/onboarding/Onboarding.tsx
@@ -126,7 +126,7 @@ export default function Onboarding({
                 }}
                 hideDefaults
                 defaultValues={{
-                  name: 'My bank account',
+                  name: 'My bank',
                   parent: accounts.type_asset as Account,
                   type: 'BANK',
                   fk_commodity: useMainCurrency().data as Commodity,
@@ -339,6 +339,7 @@ async function createInitialAccounts(setAccounts: Function, currency: Commodity)
       placeholder: true,
       fk_commodity: currency,
       parent: root,
+      hidden: true,
     }),
   ]);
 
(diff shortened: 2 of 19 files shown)


src/components/forms/price/PriceForm.tsx
```diff
@@ -95,6 +95,7 @@ export default function PriceForm({
                   placeholder="Convert to"
                   onChange={field.onChange}
                   namespace="CURRENCY"
+                  ignore={[(defaultValues?.fk_commodity as Commodity)?.guid || '']}
```
argaen, line 98: unrelated: Ignore the current commodity so we avoid users adding prices for EUR.EUR and so on

src/components/onboarding/Onboarding.tsx
```diff
@@ -339,6 +339,7 @@ async function createInitialAccounts(setAccounts: Function, currency: Commodity)
       placeholder: true,
       fk_commodity: currency,
       parent: root,
+      hidden: true,
```
argaen, line 342: hide equity account as we don't use it for anything other than opening balances and may confuse users

src/components/pages/accounts/AccountsTable.tsx
```diff
@@ -67,19 +67,23 @@ function getTreeTotals(
...
-    (total, [monthYear, amount]) => {
-      if (DateTime.fromFormat(monthYear, 'MM/yyyy') <= selectedDate) {
-        return total.add(amount);
-      }
-      return total;
-    },
-    new Money(0, current.commodity?.mnemonic || ''),
-  );
+  let accountTotal = (monthlyTotals[current.guid] || {})[selectedDate.toFormat('MM/yyyy')]
+    || new Money(0, current.commodity?.mnemonic || '');
```
argaen, line 75: monthlyTotals contains the total already for Asset and Liability accounts

src/components/pages/accounts/NetWorthPie.tsx
```diff
@@ -20,26 +20,8 @@ export default function NetWorthPie({
...
-    (total, [monthYear, amount]) => {
-      if (DateTime.fromFormat(monthYear, 'MM/yyyy') <= selectedDate) {
-        return total.add(amount);
-      }
-      return total;
-    },
-    new Money(0, unit),
-  ).toNumber() * -1;
+  const assetsTotal = assetsSeries?.[selectedDate.toFormat('MM/yyyy')] || new Money(0, unit);
+  const liabilitiesTotal = liabilitiesSeries?.[selectedDate.toFormat('MM/yyyy')] || new Money(0, unit);
```
argaen, line 24: no need to calculate totals as now this series have the total already calculated

PR #616, opened 2024-02-08
Move to react-query (first iteration)

argaen:
This is the beginning of a migration from `swr` to `react-query`. The main reason is that react-query gives you way more options on behavior and `swr` didn't do what we expected it to do sometimes:

> Using global mutator only with the key parameter will not update the cache or trigger revalidation unless there is a mounted SWR hook using the same key.

In `react-query` using `setQueryData` we got it covered.

This is not just a migration but a change on how we update the cache/ui/state. Before, we had the mutates in the Form of the entity but this creates a few problems:

- The form needs to know which entities are related
- The form needs to know which entities have changed and update accordingly
- Sometimes we need to change behavior depending on where we show the form

I've decided to make state/caching more integrated into our entities management and thus, we are overriding the `save` and `remove` methods of Account so that it updates the `/api/accounts` needed keys. This now means now, we make sure we always update state and remove eventual consistency issues.

This PR only does this for `useAccounts` and `useAccount`. Next PRs will cover other entities.


src/components/forms/transaction/TransactionForm.tsx
```diff
@@ -150,7 +150,6 @@ async function onSubmit(data: FormValues, action: 'add' | 'update' | 'delete', o
 
   transaction.splits.forEach(split => {
     if (split.account.commodity.namespace !== 'CURRENCY') {
-      mutate('/api/investments', undefined);
```
argaen, line 153: This is the main benefit, now this key is updated automatically by just calling the mutate below.

src/app/dashboard/layout.tsx
```diff
@@ -2,24 +2,34 @@
...
+    },
+  },
+});
+
 export default function DashboardLayout({
   children,
 }: React.PropsWithChildren): JSX.Element {
   const router = useRouter();
   const { data: theme } = useTheme();
-  const hookData = useDataSource();
```
argaen, line 22: move into `DashboardPage`

src/book/entities/Account.ts
```diff
@@ -68,6 +72,7 @@ export default class Account extends BaseEntity {
     type!: string;
 
   @BeforeInsert()
+  @BeforeUpdate()
```
argaen, line 75: update path on updates

src/book/entities/Account.ts
```diff
@@ -136,6 +141,77 @@ export default class Account extends BaseEntity {
...
     // transactions, they are just parents of other accounts
     placeholder!: boolean;
+
+  async save(options?: SaveOptions): Promise<this> {
+    const account = await super.save(options);
+
+    const queryClient = Account.getRepository().manager.connection.options.extra?.queryClient;
+    if (queryClient) {
+      updateCache({ queryClient, account: this });
+    }
```
argaen, line 151: fail gracefully for now but at some point we probably want to make this mandatory as without a queryClient, the frontend does not work

src/components/buttons/FormButton.tsx
```diff
@@ -43,7 +43,8 @@ export default function FormButton({
             React.cloneElement(
               children,
               {
-                onSave: () => {
+                onSave: (e: unknown) => {
+                  children.props.onSave?.(e);
```
argaen, line 47: bug as if you overrode `onSave` it didn't work. It does now

src/components/forms/account/AccountForm.tsx
```diff
@@ -302,64 +298,36 @@ export default function AccountForm({
...
     }
-  } else if (action === 'update') {
-    await account.save();
-    mutate(`/api/accounts/${account.guid}`);
-    mutate('/api/accounts');
   } else if (action === 'delete') {
-    await Account.remove(account);
-    mutate(`/api/accounts/${account.guid}`);
-    mutate('/api/accounts');
-    router.replace('/dashboard/accounts');
```
argaen, line 333: this replace has been moved out, in the page that renders the button because we won't always have to send the user to the home page.

src/components/forms/account/AccountForm.tsx
```diff
@@ -302,64 +298,36 @@ export default function AccountForm({
...
-          Account.findOneByOrFail({ guid: equityAccount?.parent.guid }),
-        ]);
-        return {
-          ...accounts,
-          [equity.guid]: equity,
-          [equityRoot.guid]: equityRoot,
-        };
-      },
-      { revalidate: false },
-    );
```
argaen, line 362: woho!

src/components/pages/accounts/AccountsTable.tsx
```diff
@@ -66,7 +66,7 @@ function getTreeTotals(
   const leaves: AccountsTableRow[] = [];
   current.childrenIds.forEach(childId => {
     const childAccount = accounts[childId];
-    if (!childAccount.hidden) {
+    if (!childAccount.hidden && childAccount.parentId === current.guid) {
```
argaen, line 69: this is kind of a hack. The childrenIds is inconsistent for when we update the parent of an account as we don't remove it from the list of childrenIds of the old parent. If we want to set the correct behavior, we need to fetch the original account and do some extra operations. Going pragmatic for now

src/hooks/useDataSource.ts
```diff
@@ -46,11 +47,12 @@ export default function useDataSource(): DataSourceContextType {
   const { storage } = useBookStorage();
   const { cache } = useSWRConfig();
   const [isLoaded, setIsLoaded] = React.useState(DATASOURCE.isInitialized);
+  const queryClient = useQueryClient();
 
   React.useEffect(() => {
     async function load() {
       let rawBook: Uint8Array;
-      if (storage && !DATASOURCE.isInitialized) {
+      if (storage && !DATASOURCE.isInitialized && queryClient) {
```
argaen, line 55: queryClient is a hard dependency now

src/layout/DashboardPage.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+import { useDataSource, DataSourceContext } from '@/hooks';
+import Loading from '@/components/Loading';
+import Topbar from '@/layout/Topbar';
+
+export default function DashboardPage({
+  children,
+}: React.PropsWithChildren): JSX.Element {
```
argaen, line 9: Created this component so we have the context for QueryClient outside in the layout and here we make sure that `useDataSource` can call `useQueryClient`.

PR #622, opened 2024-02-13
Migrate commodity to react-query

argaen:
Now `BaseEntity` has the behavior of by default saving `Account` and `Commodity` entities into their respective list and detail cache keys.

In order to do that, I modified `useAccounts` so it returns `Account[]` instead of an `AccountsMap`. Now the components that need the map create it themselves.

The next entities are going to be a bit more complicated as we don't work with list/detail so they'll probably need a custom `updateCache` method.


src/book/entities/Account.ts
```diff
@@ -145,73 +145,70 @@ export default class Account extends BaseEntity {
...
     return account;
   }
 }
 
+/**
+ * Because accounts have parent/children relations,
+ * whenever an account is updated or deleted we need to update
+ * those relations
+ */
 export async function updateCache(
```
argaen, line 170: this now contains only behavior about updating parent children as the "normal" behavior is in BaseEntity

PR #623, opened 2024-02-13
Move useStartDate and useLatestTxs to react query

argaen:
Given these queries are very low cost, we are taking the approach of invalidating them whenever data changes rather than pre-computing them


src/book/entities/Transaction.ts
```diff
@@ -75,6 +77,50 @@ export default class Transaction extends BaseEntity {
...
+    queryClient,
+  }: {
+    queryClient: QueryClient,
+    entity: Transaction,
+    isDelete?: boolean,
+  },
+) {
+  queryClient.invalidateQueries({
+    queryKey: [Transaction.CACHE_KEY, { name: 'latest' }],
+    refetchType: 'all',
```
argaen, line 118: we need to set `all` because the queries may be inactive or active, we don't know here.

PR #624, opened 2024-02-13
Move main currency to query

argaen:


src/hooks/api/index.ts
```diff
@@ -32,12 +32,11 @@ export function useStartDate(): UseQueryResult<DateTime> {
...
 
-export function useMainCurrency(): SWRResponse<Commodity> {
-  const key = '/api/main-currency';
-  return useSWRImmutable(
-    key,
-    fetcher(queries.getMainCurrency, key),
-  );
+export function useMainCurrency(): UseQueryResult<Commodity> {
+  return useQuery({
+    queryKey: [Commodity.CACHE_KEY, { guid: 'main' }],
```
argaen, line 37: At some point we probably could re-set the main currency guid to be `main` and use always `useCommodity('main')`.

PR #626, opened 2024-02-14
Move usePrices to query

argaen:
Started trying to optimise the `usePrices` query by modifying the cached data but then realizing it's a bit more complicated due to using `upserts` for inserting new prices which means we have to check via `guid` but also via the conflict keys when upserting. For now taking the strategy of invalidating the keys.

### DIFF
--- a/src/app/dashboard/commodities/[guid]/page.tsx
+++ b/src/app/dashboard/commodities/[guid]/page.tsx
@@ -2,6 +2,7 @@
 
 import React from 'react';
 import { BiEdit, BiPlusCircle } from 'react-icons/bi';
+import Link from 'next/link';
 
 import { useCommodity, usePrices } from '@/hooks/api';
 import FormButton from '@/components/buttons/FormButton';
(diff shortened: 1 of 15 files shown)


src/app/dashboard/commodities/[guid]/page.tsx
```diff
@@ -102,7 +103,14 @@ export default function CommodityPage({ params }: CommodityPageProps): JSX.Eleme
...
                   <span className="title text-lg">
-                    {`${commodity.mnemonic} - ${byCurrencyPrices[0].currency.mnemonic} price data`}
+                    {`${commodity.mnemonic} - `}
+                    <Link
+                      href={`/dashboard/commodities/${byCurrencyPrices[0].currency.guid}`}
+                    >
+                      {byCurrencyPrices[0].currency.mnemonic}
+                    </Link>
+                    {' '}
+                    price data
```
Comment (argaen) [inline src/app/dashboard/commodities/[guid]/page.tsx:113]: Unrelated but useful, adding a link to the currency so can jump directly

src/book/entities/Price.ts
```diff
@@ -94,6 +98,64 @@ export default class Price extends BaseEntity {
...
     return `${this.date.toISODate()}.${this.commodity.mnemonic}.${this.currency.mnemonic}`;
   }
+
+  async save(): Promise<this> {
+    await Price.upsert(
+      [this],
+      {
+        conflictPaths: ['fk_commodity', 'fk_currency', 'date'],
+      },
+    );
```
argaen, line 108: instead of having to use `upsert` in forms or other parts where we have to remember that, now we just make sure that any save creates an upsert

PR #630, opened 2024-02-14
Move use investments to query

argaen:
This one was a bit more complex compared to the previous ones as we've modified the current hook to make use of dependant data so whenever there are updates, the investment gets updated accordingly too.

### DIFF
--- a/src/app/dashboard/investments/page.tsx
+++ b/src/app/dashboard/investments/page.tsx
@@ -15,11 +15,11 @@ import Money from '@/book/Money';
 import * as API from '@/hooks/api';
 
 export default function InvestmentsPage(): JSX.Element {
-  const { data: investments, isLoading } = API.useInvestments();
+  const { data: investments, isPending } = API.useInvestments();
   const { data: currency } = API.useMainCurrency();
   const mainCurrency = currency?.mnemonic || 'EUR';
 
-  if (isLoading) {
+  if (isPending) {
     return (
       <div className="h-screen">
         <div className="flex text-sm h-3/4 place-content-center place-items-center">
--- a/src/app/dashboard/layout.tsx
+++ b/src/app/dashboard/layout.tsx
@@ -20,7 +20,7 @@ const queryClient = new QueryClient({
       refetchOnMount: false,
       refetchOnReconnect: false,
       refetchOnWindowFocus: false,
-      gcTime: 3600000,
+      gcTime: 300000,
     },
   },
 });
--- a/src/hooks/api/useInvestments.ts
+++ b/src/hooks/api/useInvestments.ts
@@ -1,0 +1,26 @@
+import {
+  useQuery,
+  UseQueryResult,
+} from '@tanstack/react-query';
+
+import { InvestmentAccount } from '@/book/models';
+import { getInvestment, getInvestments } from '@/lib/queries';
+import fetcher from './fetcher';
+
+export function useInvestment(guid: string): UseQueryResult<InvestmentAccount> {
+  const result = useQuery({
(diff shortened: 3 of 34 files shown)


src/app/dashboard/layout.tsx
```diff
@@ -20,7 +20,7 @@ const queryClient = new QueryClient({
       refetchOnMount: false,
       refetchOnReconnect: false,
       refetchOnWindowFocus: false,
-      gcTime: 3600000,
+      gcTime: 300000,
```
argaen, line 23: reducing to 5 minutes to clear up memory more often as we have orphaned keys over time due to how react-query recommends to depend on other data (by creating new keys)

src/app/dashboard/investments/page.tsx
```diff
@@ -15,11 +15,11 @@ import Money from '@/book/Money';
 import * as API from '@/hooks/api';
 
 export default function InvestmentsPage(): JSX.Element {
-  const { data: investments, isLoading } = API.useInvestments();
+  const { data: investments, isPending } = API.useInvestments();
   const { data: currency } = API.useMainCurrency();
   const mainCurrency = currency?.mnemonic || 'EUR';
 
-  if (isLoading) {
+  if (isPending) {
```
argaen, line 22: `isPending` is the recommended approach as we don't want to show `loading` when we fetch data from other keys

src/hooks/api/useInvestments.ts
```diff
@@ -0,0 +1,115 @@
...
+  const { data: mainCurrency, dataUpdatedAt: currencyUpdatedAt } = useMainCurrency();
+
+  const result = useQuery({
+    queryKey: [
+      InvestmentAccount.CACHE_KEY,
+      {
+        account: guid,
+        accountUpdatedAt,
+        splitsUpdatedAt,
+        currencyUpdatedAt,
```
argaen, line 48: using the updatedAt instead of the full object for memory usage sake

src/hooks/api/useInvestments.ts
```diff
@@ -0,0 +1,115 @@
...
+            return newEntities;
+          },
+        );
+
+        return investment;
+      },
+      `${InvestmentAccount.CACHE_KEY}/${guid}`,
+    ),
+    enabled: !!account && !!splits && !!mainCurrency,
+    placeholderData: keepPreviousData,
```
argaen, line 83: this avoids having glitches for when we load a different key. While loading, it keeps the previous data.

src/hooks/api/useInvestments.ts
```diff
@@ -0,0 +1,115 @@
...
+      },
+    ],
+    queryFn: fetcher(
+      async () => {
+        const investment = await initInvestment(
+          account as Account,
+          mainCurrency as Commodity,
+          splits as Split[],
+        );
+        queryClient.setQueryData(
```
argaen, line 58: whenever we create/update a new investment, make sure we update it in the `/api/investments` too. This is the only way of updating as InvestmentAccount is not an entity with a `save` method.

src/hooks/api/useSplits.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Split } from '@/book/entities';
+import type { Account } from '@/book/entities';
+import { getSplits } from '@/lib/queries';
+import fetcher from './fetcher';
+
+/**
+ * This query is mainly used for when we want to retrieve splits for a specific account
+ * so we display them
+ */
+export function useSplits(findOptions: FindOptionsWhere<Account>): UseQueryResult<Split[]> {
```
argaen, line 16: Make it generic so we can pass FindOptions

PR #631, opened 2024-02-14
Fix investment splits ordering

argaen:
We changed the ordering of the splits and didn't reflect as expected in `InvestmentAccount` which caused errors for stocks that had Sell operations. This fixes it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@fontsource/inter": "^5.0.16",
     "@hookform/resolvers": "^3.3.4",
     "@tanstack/react-query": "^5.20.5",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "@fontsource/inter": "^5.0.16",
     "@hookform/resolvers": "^3.3.4",
     "@tanstack/react-query": "^5.20.5",
-    "@tanstack/react-table": "^8.11.8",
+    "@tanstack/react-table": "^8.12.0",
```
argaen, line 27: unrelated

PR #633, opened 2024-02-15
Move monthly totals to query

argaen:
This is the final migration from SWR to react-query for data. Only the state is missing which should be quite quick.

Note we've removed the `preloading` of some data we had in the datasource hook so things would load faster because with react-query it doesn't affect. Overall though, first load is a bit slower. I'll be working on this by adding some pagination and moving queries to more granular components. Right now the monthly-totals is way too fat as we load all the history. Some ideas to improve:

- We only need global totals for accounts tree, not per month. This should depend on the date the user selects and the query should be easy to do. The problem comes a bit for nested accounts as we have to accumulate those totals
- For graphs, we only need to show latest 6 months rather than ALL. We could load the rest dynamically whenever the user pans (or disable it and see a better way)

### DIFF
--- a/src/hooks/api/usePrices.ts
+++ b/src/hooks/api/usePrices.ts
@@ -10,10 +10,20 @@ import fetcher from './fetcher';
 /**
  * Returns prices for a given commodity
  */
-export function usePrices(c: Commodity | undefined): UseQueryResult<PriceDBMap> {
+export function usePrices(params: {
+  from?: Commodity,
+  to?: Commodity,
+}): UseQueryResult<PriceDBMap> {
   return useQuery({
-    queryKey: [Price.CACHE_KEY, { commodity: c?.guid }],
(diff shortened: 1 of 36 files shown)


src/hooks/api/usePrices.ts
```diff
@@ -10,10 +10,20 @@ import fetcher from './fetcher';
...
-    enabled: !!c,
+    queryKey: [Price.CACHE_KEY, { from: params.from?.guid, to: params.to?.guid }],
+    queryFn: fetcher(
+      () => getPrices(params),
+      `${Price.CACHE_KEY}/${params.from?.guid}.${params.to?.guid}`,
+    ),
+    enabled: (
+      !!('from' in params && params.from)
+      || !!('to' in params && params.to)
+      || (!('from' in params) && !('to' in params))
```
argaen, line 26: this is to make sure we don't trigger the query when we pass from or to explicitly but with `undefined` value. This usually means the commodity hasn't been loaded yet

PR #634, opened 2024-02-15
Dont block on stocker price

argaen:
Avoid blocking on stocker prices by invalidating prices key once we've inserted prices.

Also, refactored the way we define the keys so we can invalidate/refetch groups of keys accordingly. More in https://stackoverflow.com/questions/74770055/react-query-invalidatequeries-partial-match-not-working

### DIFF
--- a/src/hooks/useDataSource.ts
+++ b/src/hooks/useDataSource.ts
@@ -125,10 +125,11 @@ async function importBook(storage: BookStorage, rawData: Uint8Array) {
   await DATASOURCE.sqljsManager.loadDatabase(rawBook);
 
   DATASOURCE.options.extra.queryClient.refetchQueries({
+    queryKey: ['api'],
     type: 'all',
   });
 
-  await loadStockerPrices();
+  loadStockerPrices();
   await save(storage);
 }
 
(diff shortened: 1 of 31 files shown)


src/hooks/useDataSource.ts
```diff
@@ -125,10 +125,11 @@ async function importBook(storage: BookStorage, rawData: Uint8Array) {
   await DATASOURCE.sqljsManager.loadDatabase(rawBook);
 
   DATASOURCE.options.extra.queryClient.refetchQueries({
+    queryKey: ['api'],
```
argaen, line 128: unrelated but now that we have correct syntax for keys, we can be more specific

src/lib/Stocker.ts
```diff
@@ -82,6 +82,13 @@ export async function insertTodayPrices(): Promise<void> {
     },
   );
 
+  if (prices.length) {
+    prices[0]?.queryClient?.invalidateQueries({
+      queryKey: ['api', 'prices'],
+      refetchType: 'all',
+    });
+  }
```
argaen, line 90: If we have Stocker prices, invalidate `/api/prices`

PR #635, opened 2024-02-15
Splits pagination for Transactions table

argaen:
Well this was more effort that I thought. Working on this I uncovered some (expected) bottlenecks and my bias towards pre-loading data (I guess backend experience) and inject it as dependency to the components. Solved pagination and started changing some code so we load only the data we need.

For example, now transactions table (which is splits table really) only has splits. For the cells that need a transaction, we load the transaction using `useQuery`. There's probably a sweet spot between number of transactions and big chunks of data, I guess we will discover over time. One lesson I learned in the past, it's better to do fat queries than launching hundreds of requests BUT, with fat queries we have a high limit I guess.

Anyway, I'll be changing some graphs too so they load only 6 months data rather than all of it. Whenever user selects a new date then we can load those. Same for calculating accounts totals, instead of loading splits and monthly aggregations, we can just have an sql query that calculates that.


src/app/dashboard/layout.tsx
```diff
@@ -17,8 +17,9 @@ Modal.setAppElement('#modals');
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
+      retry: false,
       staleTime: Infinity,
-      refetchOnMount: false,
+      refetchOnMount: true,
```
argaen, line 22: okay so I just discovered this is key. Instead of having to mark all invalidations with `refetchType: all` we can just invalidate (so the active ones refetch) and the non active ones will refetch once you remount the component that uses it. Quite neat and avoids doing to much computation on changes.

src/components/pages/account/AccountInfo.tsx
```diff
@@ -15,66 +14,23 @@ export type AccountInfoProps = {
...
-    }
-
-    let { quantity } = split;
-    if (account.type === 'INCOME') {
-      quantity = -quantity;
-    }
-
-    totalThisYear += quantity;
-    return true;
-  });
```
argaen, line 57: I'm removing some useless information. We will be adding extra stuff for each account type in later PRs same as I did for Investment type

src/components/pages/account/TransactionsTable.tsx
```diff
@@ -86,98 +107,75 @@ const columns: ColumnDef<Split>[] = [
...
   },
 ];
 
-function FromToAccountPartial(
-  accounts: AccountsMap,
-) {
-  return function FromToAccount({ row }: CellContext<Split, unknown>): JSX.Element {
-    const { splits } = row.original.transaction;
-    const otherSplits = splits.filter(split => split.account.guid !== row.original.account.guid);
+function useTransaction(guid: string): UseQueryResult<Transaction> {
```
argaen, line 114: So this is quite cool. Instead of having to pre-load a bunch of data so I have all transactions available when I need them, we load the data dynamically using this hook for the Cells that need it.

src/hooks/api/useSplits.ts
```diff
@@ -1,34 +1,134 @@
...
+ * it is used by useInvestment/s queries and they need A LOT of nested information
+ * like full transation with all splits, the account, etc. We probably should separate
+ * in two queries in the future.
  */
-export function useSplits(findOptions: FindOptionsWhere<Account>): UseQueryResult<Split[]> {
-  return useQuery({
-    queryKey: [...Split.CACHE_KEY, findOptions],
+export function useSplits(
+  findOptions: FindOptionsWhere<Account>,
+): UseQueryResult<Split[]> {
```
argaen, line 20: few changes here, we now have 4 hooks in useSplits:

- useSplits: generic one, allows us to query all splits for an account, all splits for a specific account type, etc.
- useSplitsPagination: retrieve splits for an account with pagination
- useSplitsCount: get total number of splits for an account (useful for pagination to know number of pages).
- useSplitsTotal: get the total quantity for the splits of an account (useful for counting totals).

PR #637, opened 2024-02-16
Remove swr

argaen:
Move the last usages of swr (state) to react-query and remove the library completely!

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "reflect-metadata": "^0.2.1",
     "sharp": "^0.33.2",
     "sql.js": "^1.10.2",
-    "swr": "^2.2.4",
     "typeorm": "^0.3.20",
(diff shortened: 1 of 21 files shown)


package.json
```diff
@@ -58,7 +58,6 @@
     "reflect-metadata": "^0.2.1",
     "sharp": "^0.33.2",
     "sql.js": "^1.10.2",
-    "swr": "^2.2.4",
```
argaen, line 61: wohoo!

src/layout/RootLayout.tsx
```diff
@@ -0,0 +1,66 @@
+'use client';
```
argaen, line 1: nothing new, just exported this so we can use QueryClient at the root layout which requires `use client` while we have the metadata in `layout.tsx`. This is so we can have queryclient for both dashboard and user pages

PR #640, opened 2024-02-16
Set hardcoded date for demo

argaen:
Set hardcoded date to 2024-01-01 so we show proper dashboards with the hardcoded data we have.


src/app/dashboard/layout.tsx
```diff
@@ -23,6 +23,7 @@ const queryClient = new QueryClient({
       refetchOnReconnect: false,
       refetchOnWindowFocus: false,
       gcTime: 300000,
+      placeholderData: keepPreviousData,
```
argaen, line 26: keepPrevious data by default as we want to avoid flickering of UI. Once we implement skeleton UIs it won't be necessary for some of the components but it's a good default for now.

src/components/DateRangeInput.tsx
```diff
@@ -20,7 +20,7 @@ export default function DateRangeInput({
 }: DateRangeInputProps): JSX.Element {
   const now = DateTime.now();
   const shortcuts: { [key: string]: { text: string, period: { start: Date, end: Date } } } = {
-    today: {
+    t: {
```
argaen, line 23: apparently having the `today` key has some magic behavior for the datepicker where it ignores the date you pass for setting the calendar.

PR #642, opened 2024-02-19
Minor updates

argaen:
Some minor changes including:

- Fix bug where startDate for investment charts is wrongly calculated (due to `reverse` mutation of splits)
- Change account selector to display color code according to account type
- Sort accounts in account selector
- Show focused/active element in account selector (some css priority issue was avoiding that so even though you could select with arrows, you didn't have any visual feedback).
- Now main currency is selected by default in AccountForm when no default `fk_commodity` is passed.
- mainCurrency and theme keys have now an Infinity gcTime as they never change (unless user acts on it which we manually change then in the case of theme)
- Remove Root account from account selector. We want users creating accounts within the standard Assets/L/I/E accounts

### DIFF
--- a/src/app/dashboard/layout.tsx
+++ b/src/app/dashboard/layout.tsx
@@ -22,7 +22,6 @@ const queryClient = new QueryClient({
       refetchOnMount: true,
       refetchOnReconnect: false,
       refetchOnWindowFocus: false,
-      gcTime: 300000,
       placeholderData: keepPreviousData,
     },
   },
--- a/src/book/models/InvestmentAccount.ts
+++ b/src/book/models/InvestmentAccount.ts
@@ -161,7 +161,7 @@ export default class InvestmentAccount {
     this.realizedProfit = new Money(0, this.currency);
     this.dividends.splice(0, this.dividends.length);
 
-    this.splits.reverse().filter(
+    [...this.splits].reverse().filter(
       split => split.transaction.date <= (date || DateTime.now()),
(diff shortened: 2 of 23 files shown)


src/app/dashboard/layout.tsx
```diff
@@ -22,7 +22,6 @@ const queryClient = new QueryClient({
       refetchOnMount: true,
       refetchOnReconnect: false,
       refetchOnWindowFocus: false,
-      gcTime: 300000,
```
argaen, line 25: this is a default already

src/book/models/InvestmentAccount.ts
```diff
@@ -161,7 +161,7 @@ export default class InvestmentAccount {
     this.realizedProfit = new Money(0, this.currency);
     this.dividends.splice(0, this.dividends.length);
 
-    this.splits.reverse().filter(
+    [...this.splits].reverse().filter(
```
argaen, line 164: this was mutating the splits order :/

src/components/selectors/AccountSelector.tsx
```diff
@@ -46,8 +49,24 @@ export default function AccountSelector(
       id={id}
       getOptionLabel={(option: Account) => option.path}
       getOptionValue={(option: Account) => option.path}
-      options={options}
+      options={options.sort((a, b) => a.path.localeCompare(b.path))}
```
argaen, line 52: sort alphabetically

PR #644, opened 2024-02-19
Income/Expense account mainCurrency only

argaen:
Sooo this PR can be a bit controversial, I'm removing the feature of being able to have income/expense accounts in a commodity different than mainCurrency. Some reasoning here:

- When you check income/expense reports you want to see them in your main currency.
- With the current feature, it creates a bit of duality where some accounts that already existed are in EUR but some new accounts I created in SGD after I moved there. However, it's weird to have one of each and for consistency, I prefer all of them to be in EUR.

There is one use case where I think this would make sense and would be when you permanently move to another country. For this case (although not implemented yet) the recommendation anyway is to close and open a new book.

### DIFF
--- a/src/layout/RootLayout.tsx
+++ b/src/layout/RootLayout.tsx
@@ -10,7 +10,6 @@ import '@/css/globals.css';
 import { Auth0Provider } from '@/lib/auth0-provider';
 import { isProd, isStaging } from '@/helpers/env';
 
-Settings.throwOnInvalid = true;
 if (isStaging()) {
   Settings.now = () => 1704067200000; // 2023-01-01
 }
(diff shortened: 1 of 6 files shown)


src/layout/RootLayout.tsx
```diff
@@ -10,7 +10,6 @@ import '@/css/globals.css';
 import { Auth0Provider } from '@/lib/auth0-provider';
 import { isProd, isStaging } from '@/helpers/env';
 
-Settings.throwOnInvalid = true;
```
argaen, line 13: unrelated but just discovered this causes exceptions when in transaction form you type invalid year (i.e. 5 digits).

PR #649, opened 2024-02-20
Refactor total aggregations

argaen:
getMonthlyTotals is quite costly as we retrieve all splits aggregating monthly. I'm adding a new total aggregation that just aggregates all splits without the monthly aggregation. This is in direction of trying to have less heavy hooks when loading the page.

In later PRs I'll change getMonthlyTotals so it loads only the visible months (i.e. 6 or so).

I've also changed the `AccountsTable` a bit so it gets account ids passed. This will allow to display the accounts tree anywhere (placeholder accounts to show the children for example :)).


src/components/pages/account/AccountInfo.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import StatisticsWidget from '@/components/StatisticsWidget';
 import type { Account } from '@/book/entities';
 import Money from '@/book/Money';
-import { useSplitsTotal } from '@/hooks/api';
+import { useAccountTotal } from '@/hooks/api';
```
argaen, line 6: no change, just renamed

src/hooks/api/useSplits.ts
```diff
@@ -130,3 +136,41 @@ export function useSplitsTotal(
...
+  const { data: accounts, dataUpdatedAt: accountsUpdatedAt } = useAccounts();
+  const { data: prices, dataUpdatedAt: pricesUpdatedAt } = usePrices({});
+
+  const queryKey = [
+    ...Split.CACHE_KEY,
+    {
+      aggregation: 'total',
+      date: selectedDate.toISODate(),
+      accountsUpdatedAt,
+      pricesUpdatedAt,
```
argaen, line 159: need this to update when the dependencies change

PR #654, opened 2024-02-21
Fix accounts total calculation

argaen:
There was a misalignment between networth histogram and the absolute totals in the networth pie. This is because the selected date was not being used to convert the quantities when calculating absolute totals.

### DIFF
--- a/src/app/dashboard/accounts/page.tsx
+++ b/src/app/dashboard/accounts/page.tsx
@@ -16,15 +16,14 @@ import {
 import Loading from '@/components/Loading';
 import DateRangeInput from '@/components/DateRangeInput';
 import Onboarding from '@/components/onboarding/Onboarding';
-import * as API from '@/hooks/api';
+import { useAccounts } from '@/hooks/api';
 import mapAccounts from '@/helpers/mapAccounts';
 
 const NetWorthHistogram = dynamic(() => import('@/components/pages/accounts/NetWorthHistogram'), { ssr: false });
 const IncomeExpenseHistogram = dynamic(() => import('@/components/pages/accounts/IncomeExpenseHistogram'), { ssr: false });
 
 export default function AccountsPage(): JSX.Element {
-  const { data: earliestDate } = API.useStartDate();
-  const { data, isLoading } = API.useAccounts();
+  const { data, isLoading } = useAccounts();
   const [selectedDate, setSelectedDate] = React.useState(DateTime.now());
 
   if (isLoading) {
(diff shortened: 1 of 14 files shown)


src/app/dashboard/accounts/page.tsx
```diff
@@ -49,7 +48,6 @@ export default function AccountsPage(): JSX.Element {
         </span>
         <span className="ml-auto mr-3">
           <DateRangeInput
-            earliestDate={earliestDate}
```
argaen, line 52: Moved inside the component since it's the only one using it

src/components/pages/accounts/IncomeExpenseHistogram.tsx
```diff
@@ -15,7 +15,7 @@ export default function IncomeExpenseHistogram({
 }: IncomeExpenseHistogramProps): JSX.Element {
   const interval = Interval.fromDateTimes(
     selectedDate.minus({ months: 6 }).startOf('month'),
-    selectedDate.endOf('month'),
+    selectedDate,
```
argaen, line 18: don't use endOf('month') so we can check quantities middle of the month in the histogram. A big plus is also that then the quantities in the histogram and in the accounts table do match

src/components/pages/accounts/MonthlyTotalHistogram.tsx
```diff
@@ -20,15 +20,14 @@ export default function MonthlyTotalHistogram({
 }: MonthlyTotalHistogramProps): JSX.Element {
   const interval = Interval.fromDateTimes(
     selectedDate.minus({ months: 6 }).startOf('month'),
-    selectedDate.endOf('month'),
+    selectedDate,
```
argaen, line 23: same as above

src/lib/queries/getAccountsTotals.ts
```diff
@@ -86,13 +89,15 @@ function convertToParentCommodity(
...
   prices: PriceDBMap,
+  selectedDate: DateTime,
 ): Money {
   let rate = 1;
   let currency = from;
 
   if (currency.namespace !== 'CURRENCY') {
     const stockPrice = prices.getInvestmentPrice(
       currency.mnemonic,
+      selectedDate,
```
argaen, line 100: yep, this is the fix that was missing

src/lib/queries/getMonthlyTotals.ts
```diff
@@ -39,9 +39,7 @@ export default async function getMonthlyTotals(
 ): Promise<MonthlyTotals> {
   const startDate = await getEarliestDate();
   interval = interval
-    // For end we set `endOf('month')` because we are doing monthly aggregation
-    // so we want to take into account the whole month
```
argaen, line 43: lie! this makes quantities to not match with absolute totals. It's ok to have partial month data because users may select to see up until 15th of a month

PR #655, opened 2024-02-21
Extract aggregate children totals

argaen:
First step for refactoring monthly worth. I've extracted the aggregation of children totals into the parent as we will use this on demand. The idea is as follows:

- useAccountsTotal will now return the total for each account without taking into account the children.
- it accepts a select parameter for when you want to add the aggregation so you display accordingly.

It's true that it is extra computing but let's see how it reacts and we can always adapt.

### DIFF
--- a/src/lib/queries/getAccountsTotals.ts
+++ b/src/lib/queries/getAccountsTotals.ts
@@ -1,10 +1,11 @@
 import { DateTime } from 'luxon';
 
 import Money from '@/book/Money';
-import { Commodity, Split } from '@/book/entities';
+import { Split } from '@/book/entities';
 import type { Account } from '@/book/entities';
 import mapAccounts from '@/helpers/mapAccounts';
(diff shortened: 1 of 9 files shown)


src/lib/queries/getAccountsTotals.ts
```diff
@@ -15,7 +16,7 @@ export default async function getAccountsTotals(
   const rows: { total: number, accountId: string, mnemonic: string }[] = await Split
     .query(`
       SELECT
-        SUM(cast(splits.quantity_num as REAL) / splits.quantity_denom) as total,
+        ABS(SUM(cast(splits.quantity_num as REAL) / splits.quantity_denom)) as total,
```
argaen, line 19: we don't gain anything from INCOME and LIABILITY accounts being negative so converting to ABS here rather than having specific logic in the components to display one way or the other.

PR #657, opened 2024-02-22
Refactor monthly aggregations

argaen:
Refactor monthly aggregations to re-use and make the code a bit more flexible.

We now have two hooks for aggregating monthly:

- useAccountsMonthlyTotal: This one is used for split histogram, to see how much you transfer/spend/earn monthly.
- useAccountsMonthlyWorth: This one accumulates the previous months. It's good to see net worth progression or evolution of accounts balance over time.


src/components/pages/accounts/IncomeExpenseHistogram.tsx
```diff
@@ -52,17 +52,6 @@ export default function IncomeExpenseHistogram({
     },
   ];
 
-  dates.forEach(date => {
-    const monthYear = (date as DateTime).toFormat('MM/yyyy');
```
argaen, line 56: the data format has changed, if we request monthly totals for 6 months then we return an array of 6 elements with `AccountsTotals` type.

src/hooks/api/useSplits.ts
```diff
@@ -184,34 +197,125 @@ export function useAccountsTotal(
...
     DateTime.now().minus({ month: 6 }).startOf('month'),
     DateTime.now(),
   );
   const { data: accounts, dataUpdatedAt: accountsUpdatedAt } = useAccounts();
   const { data: prices, dataUpdatedAt: pricesUpdatedAt } = usePrices({});
 
+  const aggregate = React.useCallback(
+    ((data: AccountsTotals[]) => {
+      const dates = (interval as Interval).splitBy({ month: 1 }).map(d => (d.start as DateTime).endOf('month'));
+      dates[dates.length - 1] = (interval as Interval).end as DateTime;
```
argaen, line 211: we set end of month for all dates because it's a monthly histogram so we want to see the accumulated monthly EXCEPT for the last one, where the selectedDate is important (as user may only want to see up until 15th of the month for example)

src/hooks/api/useSplits.ts
```diff
@@ -184,34 +197,125 @@ export function useAccountsTotal(
...
+      DateTime.now(),
+    );
+  const { data: accounts, dataUpdatedAt: accountsUpdatedAt } = useAccounts();
+  const { data: prices, dataUpdatedAt: pricesUpdatedAt } = usePrices({});
+  const { data: totals, dataUpdatedAt: totalsUpdatedAt } = useAccountsTotals(
+    (interval.start as DateTime).endOf('month') as DateTime,
+    (data) => data,
+  );
+
+  const aggregate = React.useCallback(
```
argaen, line 271: This one is similar to the monthlyTotal above EXCEPT that for the first month, we retrieve the total for the accounts using `useAccountsTotals` and then we set it as the first month data. We get monthly totals then only for interval.start + 1 to interval.end

src/layout/RootLayout.tsx
```diff
@@ -10,6 +10,10 @@ import '@/css/globals.css';
 import { Auth0Provider } from '@/lib/auth0-provider';
 import { isProd, isStaging } from '@/helpers/env';
 
+if (process.env.NODE_ENV === 'development') {
+  Settings.throwOnInvalid = true;
+}
```
argaen, line 15: annoying to have silent failures on wrong dates while developing and will also help us catch bugs

src/lib/queries/getAccountsTotals.ts
```diff
@@ -4,23 +4,24 @@ import Money from '@/book/Money';
...
+ */
 export default async function getAccountsTotals(
   accounts: Account[],
-  prices: PriceDBMap,
   selectedDate: DateTime,
 ): Promise<AccountsTotals> {
   const rows: { total: number, accountId: string, mnemonic: string }[] = await Split
     .query(`
       SELECT
-        ABS(SUM(cast(splits.quantity_num as REAL) / splits.quantity_denom)) as total,
```
argaen, line 19: This was a bug. Some accounts does make sense to have `ABS` like Income/Expense but some accounts may have both negative or positive amounts when calculating partial totals like liabilities/assets (they have debit/credit)

PR #658, opened 2024-02-22
Show always 7 bars in histogram

argaen:
When selecting the first of the month in the date picker, we were showing only 6 bars instead of 7 which as a side effect, caused exceptions/errors when changing back to another day with 7 bars because of the keepExistingData of react-query.

This is because when selecting a date in the datepicker, it was selecting the very first second of the day which meant the `splitBy` function of Interval would return less intervals than expected.

For example, if we have `2023-01-01/2023-03-01` what we wanted is three intervals but it would return only 2. To fix that, I've changed the date picker behavior to select the end of the day the user selects so we get one more interval.

### DIFF
--- a/src/components/DateRangeInput.tsx
+++ b/src/components/DateRangeInput.tsx
@@ -61,7 +61,7 @@ export default function DateRangeInput({
       onChange={(newValue) => {
         if (newValue) {
           onChange({
-            start: DateTime.fromISO(newValue.startDate as string),
+            start: DateTime.fromISO(newValue.startDate as string).endOf('day'),
             end: DateTime.fromISO(newValue.endDate as string),
           });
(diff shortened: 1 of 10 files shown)


src/components/DateRangeInput.tsx
```diff
@@ -61,7 +61,7 @@ export default function DateRangeInput({
       onChange={(newValue) => {
         if (newValue) {
           onChange({
-            start: DateTime.fromISO(newValue.startDate as string),
+            start: DateTime.fromISO(newValue.startDate as string).endOf('day'),
```
argaen, line 64: the actual fix! this makes Interval.splitBy to include the extra interval for when we select first day of the month

PR #660, opened 2024-02-23
Fix price deletion

argaen:
Updating cache and price removal were in the wrong order causing the deleted price to still be shown to the user. This PR fixes it.

### DIFF
--- a/src/__tests__/components/pages/account/AccountInfo.test.tsx
+++ b/src/__tests__/components/pages/account/AccountInfo.test.tsx
@@ -68,7 +68,7 @@ describe('AccountInfo', () => {
       1,
       {
         className: 'mr-2',
-        title: 'This account has a total of',
+        title: 'Total',
         description: '',
         stats: '€0.00',
       },
@@ -90,7 +90,7 @@ describe('AccountInfo', () => {
       1,
       {
         className: 'mr-2',
-        title: 'This account has a total of',
+        title: 'Total',
         description: '',
         stats: '€100.00',
       },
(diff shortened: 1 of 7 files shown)


src/__tests__/components/pages/account/AccountInfo.test.tsx
```diff
@@ -90,7 +90,7 @@ describe('AccountInfo', () => {
       1,
       {
         className: 'mr-2',
-        title: 'This account has a total of',
+        title: 'Total',
```
argaen, line 93: Unrelated

src/book/entities/Price.ts
```diff
@@ -115,12 +115,13 @@ export default class Price extends BaseEntity {
...
 
   async remove(options?: SaveOptions): Promise<this> {
+    const entity = await super.remove(options);
+
     if (this.queryClient) {
-      updateCache({ queryClient: this.queryClient, entity: this, isDelete: true });
+      updateCache({ queryClient: this.queryClient, entity, isDelete: true });
     }
 
-    const price = await super.remove(options);
```
argaen, line 122: We were removing after triggering the cache update... obviously it didn't work

src/lib/queries/getPrices.ts
```diff
@@ -20,9 +20,6 @@ export default async function getPrices({
         guid: to?.guid,
       },
     },
-    order: {
-      date: 'ASC',
```
argaen, line 24: No need to order as the Table component is the one ordering

PR #666, opened 2024-02-26
Improve investment info

argaen:
Display realized profit for when positions are closed.

### DIFF
--- a/src/book/models/InvestmentAccount.ts
+++ b/src/book/models/InvestmentAccount.ts
@@ -12,6 +12,7 @@ export default class InvestmentAccount {
   readonly account: Account;
   readonly splits: Split[];
   quantity: Money;
+  totalBought: Money;
   realizedProfit: Money;
   realizedProfitInCurrency: Money;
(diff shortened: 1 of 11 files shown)


src/book/models/InvestmentAccount.ts
```diff
@@ -111,30 +108,39 @@ export default class InvestmentAccount {
...
+        (this.unrealizedProfitAbsInCurrency.toNumber() / this.costInCurrency.toNumber()) * 100
       ) + Number.EPSILON) * 100,
     ) / 100;
     return n;
   }
 
+  get realizedProfitPct(): number {
+    const n = Math.round(
+      (
+        ((this.realizedProfit.toNumber() / this.totalBought.toNumber()) * 100) + Number.EPSILON
```
argaen, line 138: This is a bit inaccurate I guess for positions that you close and re-open as it will count total bought for the stocks you haven't sold yet. Regardless, this is not shown to the user as then we would be showing unrealized profit (given the position is open).

PR #667, opened 2024-02-26
Allow to edit main split account tx form

argaen:
When creating a new account and wanting to move a bunch of transactions from one to another, it becomes quite difficult as currently you can't edit the main split account and thus, you have to go through the other account in the transaction, filtering for that specific type of transaction (because we don't have global search implemented in the transactions table)

### DIFF
--- a/src/app/dashboard/accounts/page.tsx
+++ b/src/app/dashboard/accounts/page.tsx
@@ -3,7 +3,6 @@
 import React from 'react';
 import { DateTime } from 'luxon';
 import { BiPlusCircle } from 'react-icons/bi';
-import dynamic from 'next/dynamic';
 
 import FormButton from '@/components/buttons/FormButton';
 import AccountForm from '@/components/forms/account/AccountForm';
@@ -11,6 +10,8 @@ import {
   AccountsTable,
   NetWorthPie,
   MonthlyTotalHistogram,
+  NetWorthHistogram,
+  IncomeExpenseHistogram,
   LatestTransactions,
 } from '@/components/pages/accounts';
 import Loading from '@/components/Loading';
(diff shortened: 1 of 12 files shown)


src/app/dashboard/accounts/page.tsx
```diff
@@ -19,9 +20,6 @@ import Onboarding from '@/components/onboarding/Onboarding';
 import { useAccounts } from '@/hooks/api';
 import mapAccounts from '@/helpers/mapAccounts';
 
-const NetWorthHistogram = dynamic(() => import('@/components/pages/accounts/NetWorthHistogram'), { ssr: false });
-const IncomeExpenseHistogram = dynamic(() => import('@/components/pages/accounts/IncomeExpenseHistogram'), { ssr: false });
```
argaen, line 23: unrelated but since we don't use zoom/pan in these components anymore, this is not needed.

src/components/pages/account/InvestmentChart.tsx
```diff
@@ -166,7 +166,9 @@ export default function InvestmentChart({
             border: {
               display: false,
             },
-            min: DateTime.now().minus({ year: 1 }).toISODate(),
+            min: (startDate.toMillis() > DateTime.now().minus({ year: 1 }).toMillis())
+              ? startDate.toISODate() || undefined
+              : DateTime.now().minus({ year: 1 }).toISODate() || undefined,
```
argaen, line 171: For investments that are not 1 year, just show from startDate

src/book/entities/Transaction.ts
```diff
@@ -115,6 +115,9 @@ export async function updateCache(
     isDelete?: boolean,
   },
 ) {
+  queryClient.invalidateQueries({
+    queryKey: [...Transaction.CACHE_KEY, entity.guid],
+  });
```
argaen, line 120: unrelated but bug: Due to us caching each transaction with useQuery, we want to invalidate the transaction so when we load it in the form, it shows as expected (previously it was showing old values)

PR #668, opened 2024-02-26
Execute vacuum on save

argaen:
My compressed file size went from 2.3M to 1.4M when executing this so now we execute every time before saving. It's probably a bit overkill but I don't have a better option right now other than periodically executing which adds a bit of complexity and redundancy so leaving as is for now.

### DIFF
--- a/src/hooks/useDataSource.ts
+++ b/src/hooks/useDataSource.ts
@@ -1,6 +1,7 @@
 import React from 'react';
 import { DataSource } from 'typeorm';
 import initSqlJs from 'sql.js';
+import { Settings } from 'luxon';
 import pako from 'pako';
 
 import { insertTodayPrices } from '@/lib/Stocker';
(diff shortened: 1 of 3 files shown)


src/hooks/useDataSource.ts
```diff
@@ -107,12 +108,16 @@ export default function useDataSource(): DataSourceContextType {
...
+  await DATASOURCE.query('VACUUM');
   const rawBook = DATASOURCE.sqljsManager.exportDatabase();
   await storage.save(rawBook);
   DATASOURCE.options.extra.queryClient.setQueryData(['state', 'isSaving'], false);
 }
 
 async function importBook(storage: BookStorage, rawData: Uint8Array) {
+  if (isStaging()) {
+    Settings.now = () => Date.now();
+  }
```
argaen, line 120: unrelated but this allows users to import their own files and go to any date. Without this, can only check their data up until 1st Jan 2023

src/lib/storage/GDriveBookStorage.ts
```diff
@@ -101,7 +101,7 @@ export default class GDriveBookStorage implements BookStorage {
           Authorization: `Bearer ${this.gapiClient.getToken().access_token}`,
           'Content-Type': 'application/vnd.sqlite3',
         }),
-        body: new Blob([pako.gzip(rawBook)], { type: 'application/vnd.sqlite3' }),
+        body: new Blob([pako.gzip(rawBook, { level: 9 })], { type: 'application/vnd.sqlite3' }),
```
argaen, line 104: This didn't bring any change but leaving as is just in case since default level is 6

PR #669, opened 2024-02-26
AssetInfo page

argaen:
Start of custom page for Assets. Only adding total and networth histogram for now.

### DIFF
--- a/src/components/charts/NetWorthHistogram.tsx
+++ b/src/components/charts/NetWorthHistogram.tsx
@@ -9,7 +9,7 @@ import {
 
 import Bar from '@/components/charts/Bar';
 import { moneyToString } from '@/helpers/number';
-import { useAccountsMonthlyWorth, useMainCurrency } from '@/hooks/api';
+import { useAccountsMonthlyWorth } from '@/hooks/api';
 import monthlyDates from '@/helpers/monthlyDates';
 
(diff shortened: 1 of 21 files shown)


src/components/charts/NetWorthHistogram.tsx
```diff
@@ -21,70 +21,85 @@ C.register(
...
 
 export type NetWorthHistogramProps = {
+  assetsGuid: string,
+  assetsLabel?: string,
+  hideAssets?: boolean,
+  liabilitiesGuid: string,
+  liabilitiesLabel?: string,
+  hideLiabilities?: boolean,
   selectedDate?: DateTime,
+  showLegend?: boolean,
```
argaen, line 31: reconverting to a re-usable component where you can choose to show/hide some of the datasets.

src/components/pages/account/Header.tsx
```diff
@@ -24,7 +24,7 @@ export default function Header({
   account,
 }: HeaderProps): JSX.Element {
   const { data: parent } = useAccount(account.parentId);
-  const { data: splits } = useSplits({ guid: account.guid });
+  const { data: splits } = useSplitsPagination(account.guid);
```
argaen, line 27: we only need the splits to know the latest date and if there are actual splits so using the same that is already loaded for the transactions table which reduces the amount of queries.

PR #673, opened 2024-02-27
Placeholder improvement

argaen:
Working with custom page for each asset type, we need to improve the behavior of placeholder accounts.

> Placeholder accounts are accounts that are parents to other accounts. These accounts **can't** have transactions.

The improvements done:

- When creating an account, you can only select placeholder accounts as a parent
- Placeholder accounts are now clickable as they'll show accumulated reports for the sub-accounts they contain
- Show Add transaction button from placeholder account detail page
- Add validation so users can't set an account as placeholder if it has transactions

Also as an extra, added a monthly increase/decrease description in the total widget for account detail


src/app/dashboard/accounts/[guid]/page.tsx
```diff
@@ -61,9 +61,14 @@ export default function AccountPage({ params }: AccountPageProps): JSX.Element {
...
-        <TransactionsTable account={account} />
-      </div>
+      {
+        !account.placeholder
+        && (
+          <div className="card p-0 mt-4 bg-light-100 dark:bg-dark-800">
+            <TransactionsTable account={account} />
+          </div>
+        )
+      }
```
Comment (argaen) [inline src/app/dashboard/accounts/[guid]/page.tsx:71]: Placeholders can't have transactions so we hide the transactions table

src/book/entities/Account.ts
```diff
@@ -309,3 +310,38 @@ function CheckIECommodity(validationOptions?: v.ValidationOptions) {
...
+ */
+function CheckPlaceholder(validationOptions?: v.ValidationOptions) {
+  return function f(object: Account, propertyName: string) {
+    v.registerDecorator({
+      name: 'checkPlaceholder',
+      target: object.constructor,
+      propertyName,
+      options: validationOptions,
+      validator: {
+        async validate(_: string, args: v.ValidationArguments) {
```
argaen, line 326: first async validator I think. Awesome that it works, this will enable to create better validation in the future

src/components/forms/transaction/SplitField.tsx
```diff
@@ -79,6 +79,7 @@ export default function SplitField({
                 isClearable={false}
                 isDisabled={disabled}
                 placeholder="<account>"
+                ignorePlaceholders
```
argaen, line 82: Split accounts can't be placeholders as they can't have transactions

src/components/pages/account/AssetInfo.tsx
```diff
@@ -13,9 +15,12 @@ export type AssetInfoProps = {
 export default function AssetInfo({
   account,
 }: AssetInfoProps): JSX.Element {
-  const { data: t } = useAccountTotal(account.guid);
+  const { data: t0 } = useAccountsTotals();
+  const { data: t1 } = useAccountsTotals(DateTime.now().minus({ month: 1 }).endOf('month'));
```
argaen, line 19: End of month previous total to make the difference between today and last month (increase or decrease)

src/components/pages/account/Header.tsx
```diff
@@ -43,37 +43,51 @@ export default function Header({
...
-              <>
-                <BiPlusCircle className="mr-1" />
-                Add transaction
-              </>
+        <div
+          className={classNames(
+            'flex gap-1',
+            {
+              hidden: account.parentId === 'rootAccount',
+            },
```
argaen, line 54: Main accounts (assets, expenses, etc) can't be edited so we hide the buttons

src/components/pages/account/Header.tsx
```diff
@@ -43,37 +43,51 @@ export default function Header({
...
+          )}
+        >
+          <div
+            className={classNames(
+              '',
+              {
+                hidden: account.placeholder,
+              },
```
argaen, line 62: If an account is a placeholder, don't show add transaction button

src/components/pages/accounts/AccountsTable.tsx
```diff
@@ -110,27 +110,13 @@ const columns: ColumnDef<AccountsTableRow>[] = [
...
-            </Link>
-          )
-        }
+        <Link
+          data-tooltip-id={row.original.account.guid}
+          className={accountColorCode(row.original.account, 'badge hover:text-slate-300')}
+          href={`/dashboard/accounts/${row.original.account.guid}`}
+        >
+          {row.original.account.name}
+        </Link>
```
argaen, line 119: Placeholder accounts are clickable now

src/lib/queries/getAccountsTotals.ts
```diff
@@ -21,7 +21,6 @@ export default async function getAccountsTotals(
         splits.account_guid as accountId
       FROM splits
       JOIN transactions as tx ON splits.tx_guid = tx.guid
-      JOIN accounts as account ON splits.account_guid = account.guid
```
argaen, line 24: not necessary

PR #678, opened 2024-02-28
Add sankey diagram for assets

argaen:
Show sankey diagram for current month. We only show this for non placeholder Asset accounts. Probably Liability accounts would benefit from this too but focusing on AssetInfo page for now.

A future upgrade also would be to add a date picker in the detail page and show data accordingly

Screenshot 2024-02-28 at 10 37 15 AM


src/components/charts/AssetSankey.tsx
```diff
@@ -0,0 +1,150 @@
...
+      JOIN splits ON splits.tx_guid = tx.guid
+      JOIN accounts ON splits.account_guid = accounts.guid
+      WHERE tx.guid IN (
+        SELECT DISTINCT tx_guid
+        FROM splits
+        WHERE account_guid = '${guid}'
+      )
+      AND tx.post_date >= '${selectedDate.startOf('month').toSQLDate()}'
+      GROUP BY splits.account_guid
+      HAVING SUM(cast(splits.quantity_num as REAL) / splits.quantity_denom) != 0
```
argaen, line 39: This is a useful query for counting total expense for a month in a given account. Probably will extra in the future as some statisticswidget showing Top category spending this month or something like this would be nice in asset account

src/components/forms/price/PriceForm.tsx
```diff
@@ -145,12 +145,17 @@ async function onSubmit(
...
+    guid: undefined,
   });
 
-  if (action === 'add' || action === 'update') {
+  if (action === 'add') {
     await price.save();
+  } else if (action === 'update') {
+    await price.save();
+    price.guid = data.guid;
+    await price.remove();
```
argaen, line 156: Discovered a bug where updating the date of an existing Price was failing because it said duplicated key. Seems upsert functionality is trying to add a new price when one of the elements in the unique key constraint changes. This is a bit confusing. Don't know if there is a better way but for now the approach I'm doing is to save without `guid` (so a new instance is added` and then delete the previous one

src/layout/RootLayout.tsx
```diff
@@ -15,7 +15,7 @@ if (process.env.NODE_ENV === 'development') {
 }
 
 if (isStaging()) {
-  Settings.now = () => 1704067200000; // 2023-01-01
+  Settings.now = () => 1703980800000; // 2023-12-31
```
argaen, line 18: So sankey diagrams get shown in the demo

PR #684, opened 2024-02-29
Add earn widget

argaen:


src/app/dashboard/accounts/[guid]/page.tsx
```diff
@@ -38,29 +38,19 @@ export default function AccountPage({ params }: AccountPageProps): JSX.Element {
     );
   }
 
+  let infoComponent: JSX.Element;
+  if (isInvestment(account)) {
+    infoComponent = <InvestmentInfo account={account} />;
+  } else if (isAsset(account) || isLiability(account)) {
```
Comment (argaen) [inline src/app/dashboard/accounts/[guid]/page.tsx:44]: AssetInfo also shown in Liability accounts. I'm not 100% all widgets are useful. For example liability accounts typically will not have income (unless there is some cashback/bonus things in a credit card). Leaving like this for now as other are potentially useful

src/components/pages/account/EarnWidget.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useMainCurrency, useCashFlow } from '@/hooks/api';
+import type { Account } from '@/book/entities';
+import StatisticsWidget from '@/components/StatisticsWidget';
+import { moneyToString } from '@/helpers/number';
+
+export type EarnWidgetProps = {
+  account: Account,
+};
+
+export default function EarnWidget({
```
argaen, line 13: This widget is almost a copy of SpendWidget except for some styling and filters. However separated because description field is a bit different and creates some complexity

PR #685, opened 2024-02-29
Fix pagination

argaen:


src/components/pages/account/TransactionsTable.tsx
```diff
@@ -37,7 +37,7 @@ export default function TransactionsTable({
     pageSize: 10,
   });
   const { data: splitsCount } = useSplitsCount(account.guid);
-  const { data: splits } = useSplitsPagination(account.guid);
+  const { data: splits } = useSplitsPagination(account.guid, { pageSize, pageIndex });
```
argaen, line 40: Removed this when added the defaults... 🤦

PR #686, opened 2024-02-29
Always show realized profit

argaen:
For positions that we've sold and re-bought or sold partially, it's quite annoying to not be able to see the realized profit in detail page


src/__tests__/components/pages/account/InvestmentInfo.test.tsx
```diff
@@ -144,125 +146,27 @@ describe('InvestmentInfo', () => {
...
-        stats: '€20.00',
-        description: '',
-      },
-      {},
-    );
-
-    expect(container).toMatchSnapshot();
-  });
-
-  it('renders as expected with closed position', async () => {
```
argaen, line 170: being open or closed doesn't make a difference now

PR #695, opened 2024-03-04
Fix split update

argaen:
Make sure we update Splits.CACHE_KEY/accountId for the account we are displaying for cases where user changes the mainsplit account to be a different one.

Currently, without this changes, when changing the split this happens:

Screenshot 2024-03-04 at 11 11 37 AM

This is in the my bank account. The transaction doesn't have a My bank split anymore but stays there because with current code we don't update the key for My bank account.

### DIFF
--- a/src/book/__tests__/entities/Transaction.test.ts
+++ b/src/book/__tests__/entities/Transaction.test.ts
@@ -286,7 +286,7 @@ describe('caching', () => {
 
     await tx.save();
 
-    expect(mockInvalidateQueries).toBeCalledTimes(8);
+    expect(mockInvalidateQueries).toBeCalledTimes(7);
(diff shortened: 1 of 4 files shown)


src/book/__tests__/entities/Transaction.test.ts
```diff
@@ -302,9 +302,6 @@ describe('caching', () => {
     expect(mockInvalidateQueries).toBeCalledWith({
       queryKey: ['api', 'splits', '2'],
     });
-    expect(mockInvalidateQueries).toBeCalledWith({
-      queryKey: ['api', 'aggregations', 'accounts', 'totals'],
-    });
```
argaen, line 307: unrelated but this key is not used anymore

PR #696, opened 2024-03-04
Fix path eventual consistency

argaen:
When updating the name of parent accounts or changing the parent's parent, the path of children was not being updated accordingly. This PR fixes this by computing this in the `useAccounts` hook instead of saving it in the DB


src/book/entities/Account.ts
```diff
@@ -71,24 +69,7 @@ export default class Account extends BaseEntity {
...
-      // against parent not existing (which is mandatory).
-      this.path = `${this.parent?.path}:${this.name}`;
-    }
-  }
-
-  @Column({
-    type: 'text',
-    default: '',
-  })
-    path!: string;
```
argaen, line 91: path is not a DB column anymore but a dynamically computed field when calling `useAccounts`.

src/book/entities/Account.ts
```diff
@@ -148,69 +129,26 @@ export default class Account extends BaseEntity {
...
-          parentAccount.childrenIds.splice(childIndex, 1);
-        }
-
-        newEntities[parentIndex] = parentAccount;
-      }
-
-      return newEntities;
-    },
-  );
+export async function updateCache({ queryClient }: { queryClient: QueryClient }) {
```
argaen, line 148: simplifying this for now as re-computing accounts is not expensive as there are not many. May re-optimise in the future but this removes complexity

src/components/forms/account/AccountForm.tsx
```diff
@@ -158,6 +158,7 @@ export default function AccountForm({
                 isClearable={false}
                 onlyPlaceholders
                 placeholder="<parent account>"
+                ignoreAccounts={[form.getValues('guid')]}
```
argaen, line 161: I've changed the purpose of `ignoreAccounts`. Initially it was to filter out types of accounts but the name is confusing and wasn't being used. Now it's with a list of guid.

This fixes so an account can't be it's own parent

src/hooks/api/useAccounts.ts
```diff
@@ -6,20 +6,44 @@ import {
 import { Account } from '@/book/entities';
 import fetcher from './fetcher';
 
-export function useAccount(guid: string): UseQueryResult<Account> {
-  const result = useQuery({
-    queryKey: [...Account.CACHE_KEY, { guid }],
-    queryFn: fetcher(() => Account.findOneBy({ guid }), `/${Account.CACHE_KEY.join('/')}/${guid}`),
-  });
+export function useAccount(guid: string): UseQueryResult<Account | undefined> {
```
argaen, line 9: instead of this being a different hook, it is now a select on top of the `useAccounts` hook. Following https://tkdodo.eu/blog/react-query-data-transformations

src/hooks/api/useAccounts.ts
```diff
@@ -6,20 +6,44 @@ import {
...
+  select?: (data: Account[]) => TData,
+): UseQueryResult<TData> {
   const result = useQuery({
     queryKey: [...Account.CACHE_KEY],
-    queryFn: fetcher(() => Account.find(), `/${Account.CACHE_KEY.join('/')}`),
+    queryFn: fetcher(
+      async () => {
+        const accounts = await Account.find();
+        const root = accounts.find(a => a.type === 'ROOT') as Account;
+        setAccountPaths(root, accounts);
```
argaen, line 26: every time we re-fetch the accounts, we calculate the paths. Yes, it's true there are chances we re-calculate the same but for now leaving as is, we can always optimise

src/book/entities/BaseEntity.ts
```diff
@@ -31,78 +30,8 @@ export default class BaseEntity extends BE {
...
-      updateCache({ queryClient: this.queryClient, entity });
-    }
-
-    return entity;
-  }
-
-  async remove(options?: SaveOptions): Promise<this> {
-    const entity = await super.remove(options);
-
-    if (this.queryClient && ['Account', 'Commodity'].includes(this.constructor.name)) {
```
argaen, line 47: While it is an optimization, it adds some complexity and potential eventual consistency problems. Account and Commodity are "cheap" entities to retrieve every time so removing for now

src/components/onboarding/Onboarding.tsx
```diff
@@ -22,7 +22,7 @@ export default function Onboarding({
   const { save, datasource } = React.useContext(DataSourceContext);
   const [run] = React.useState(show);
   const [stepIndex, setStepIndex] = React.useState(0);
-  const [accounts, setAccounts] = React.useState<{ [key: string]: Account }>({});
+  const { data: accounts } = useAccounts();
```
argaen, line 25: re-using `useAccounts`

src/components/pages/account/TransactionsTable.tsx
```diff
@@ -118,9 +118,7 @@ function useTransaction(guid: string): UseQueryResult<Transaction> {
       () => Transaction.findOne({
         where: { guid },
         relations: {
-          splits: {
-            fk_account: true,
-          },
+          splits: true,
```
argaen, line 121: We load the accounts using `useAccounts` instead of querying with the transaction (so path is calculated)

PR #709, opened 2024-03-06
Pie chart ie placeholder

argaen:
Screenshot 2024-03-06 at 3 54 24 PM

I'm actually tempted to add this to Asset/Liability placeholder accounts too. The only problem is that it is redundant with the TotalWidget total (but the increment is useful, maybe I'll repurpose the widget)


src/components/charts/TotalsPie.tsx
```diff
@@ -0,0 +1,89 @@
...
+export type TotalsPieProps = {
+  guids?: string[],
+  title: string,
+  backgroundColor?: string[],
+  selectedDate?: DateTime,
+  showTooltip?: boolean,
+  showDataLabels?: boolean,
+};
+
+export default function TotalsPie({
```
argaen, line 19: This is the networthpie re-purposed

src/constants/colors.ts
```diff
@@ -0,0 +1,36 @@
+export const ASSET = '#06B6D4';
+export const LIABILITY = '#FF6600';
+export const INCOME = '#22C55E';
+export const EXPENSE = '#EF4444';
```
argaen, line 4: would be good to find a way to export the tailwindcss color variables into this so we can re-use

PR #726, opened 2024-03-12
First approach to error handling

argaen:
Implement error handling using react-hot-toast and boundaries. No big implementation changes, just showing toasts to users.


src/app/user/login/page.tsx
```diff
@@ -16,23 +17,25 @@ export default function LoginPage(): JSX.Element {
     }
   }, [isAuthenticated, router]);
 
+  if (error) {
+    if (error.message === 'INVALID_SUBSCRIPTION') {
+      new AuthError(error.message, error.message).show();
```
argaen, line 22: we want the login page to still render when we have these kind of errors that's why we don't throw and just call show

PR #735, opened 2024-03-14
Improve import flow

argaen:
Importing was kind of broken before because the data source was being mixed and new saves didn't work. This PR fixes this plus few other things:

- In onboarding, we first let the user go with onboarding or import an already stored file
- Show an error when users try to import an invalid file


src/components/buttons/SaveButton.tsx
```diff
@@ -10,7 +10,7 @@ import classNames from 'classnames';
 export default function SaveButton(): JSX.Element {
   const { data: isSaving } = useQuery({
     queryKey: ['state', 'isSaving'],
-    placeholderData: false,
+    queryFn: () => false,
```
argaen, line 13: now that we are displaying errors, this was actually showing an error that queryFn was missing

src/components/charts/TotalsPie.tsx
```diff
@@ -43,8 +42,8 @@ export default function TotalsPie({
 
     return guids.map(guid => {
       let total = totals?.[guid] || new Money(0, unit);
-      const account = accounts?.find(a => a.guid === guid) as Account;
-      if (currency && account.commodity.guid !== currency?.guid) {
+      const account = accounts?.find(a => a.guid === guid);
+      if (account && currency && account.commodity.guid !== currency?.guid) {
```
argaen, line 46: for when an account is not found which can happen when importing a different file, initializing, etc.

src/components/forms/currency/CurrencyForm.tsx
```diff
@@ -2,17 +2,14 @@ import React from 'react';
...
 import Selector from '@/components/selectors/Selector';
 
 const resolver = classValidatorResolver(Commodity, { validator: { stopAtFirstError: true } });
 
-const CURRENCIES = [
-  { label: 'EUR' },
-  { label: 'USD' },
-  { label: 'SGD' },
-];
+const CURRENCIES = Object.keys(currencies).map(c => ({ label: c }));
```
argaen, line 12: all options from dinero.js are now available at onboarding!

src/components/onboarding/Onboarding.tsx
```diff
@@ -40,18 +42,55 @@ export default function Onboarding({
...
+              <div className="flex justify-center gap-3 mt-6">
+                <button
+                  type="button"
+                  className="btn btn-cta"
+                  onClick={() => setStepIndex(stepIndex + 1)}
+                >
+                  Show me!
+                </button>
+                <ImportButton
+                  onImport={() => setRun(false)}
```
argaen, line 72: if the user imports, we close the onboarding

PR #738, opened 2024-03-15
Add metadata, disallow robots

argaen:


src/app/dashboard/accounts/[guid]/layout.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+export const metadata = {
+  title: 'Account | Maffin',
```
Comment (argaen) [inline src/app/dashboard/accounts/[guid]/layout.tsx:4]: For some reason, the template does not apply to detail pages

PR #740, opened 2024-03-15
Add export functionality

argaen:
Let users export the file from the app directly. It's the same as copying the file from google drive but it's useful to have it built in

### DIFF
--- a/src/components/buttons/ExportButton.tsx
+++ b/src/components/buttons/ExportButton.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import { BiExport } from 'react-icons/bi';
+
+import { DataSourceContext } from '@/hooks';
+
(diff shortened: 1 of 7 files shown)


src/components/buttons/ExportButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <>
+      <button
+        id="export-button"
+        type="button"
+        disabled={!isLoaded}
+        className={className}
+        onClick={() => {
+          if (ref.current) {
+            const rawBook = datasource?.sqljsManager.exportDatabase() as Uint8Array;
+            const blob = new Blob([rawBook], { type: 'application/vnd.sqlite3' });
```
argaen, line 27: In the future we can use `pako.gzip` to compress the data but given Maffin supports importing both compressed and uncompressed data, doesn't make a difference and, this way, users can check the file in sqlite directly.

PR #744, opened 2024-03-18
Add free plan

argaen:
Add distinction between demo and free plans. Now there is free.maffin.io where we host a version without demo data which is the same as the paid version except:

- there is no persistence
- there is no automatic pricing from stocks

Also added an upgrade tooltip in the save button and improved the tooltip components and styles


src/app/user/login/page.tsx
```diff
@@ -30,7 +30,7 @@ export default function LoginPage(): JSX.Element {
       className="btn btn-primary"
       type="button"
       onClick={() => {
-        if (!isStaging()) {
+        if (!IS_DEMO_PLAN) {
```
argaen, line 33: we require login only when free or paid

src/components/tables/Table.tsx
```diff
@@ -55,7 +55,7 @@ export default function Table<T extends object = {}>(
 
   return (
     <>
-      <div className="relative overflow-hidden rounded-md">
```
argaen, line 58: this avoided tooltips to render properly so removing (having seen any side effect)

src/components/tooltips/UpgradeTooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+    >
+      <h2 className="flex font-normal items-center justify-center text-cyan-600">
+        <BiCrown className="mr-1 font-thin" />
+        Premium
+      </h2>
+      <div className="text-center">
+        Auto saving to persistent storage only available for premium users.
+      </div>
+      <Link
+        href="https://maffin.io"
```
argaen, line 22: will need to update once there is a "manage your subscription" page or a pricing table

src/helpers/env.ts
```diff
@@ -1,7 +1,43 @@
...
+    domain: 'maffin.eu.auth0.com',
+    clientId: 'cEXnN96kEP3ER2EDJjmjRW0u2MEFBUKK',
+    scopes: 'profile email https://www.googleapis.com/auth/drive.file',
+  },
+};
+
+const FREE_PLAN_CONFIG = {
+  auth0: {
+    domain: 'maffin-dev.eu.auth0.com',
+    clientId: 'mMmnR4NbQOnim9B8QZfe9wfFuaKb8rwW',
```
argaen, line 18: yes, we use dev for pricing plans. Dunno if it's temporary or not but it is to avoid breaking prod when doing changes

PR #755, opened 2024-03-20
Set global interval

argaen:
Instead of having the interval defined in the accounts paged, use a state hook with react-query to set it. This is because this interval is reused in mostly all the graphs and want to persist it between page changes for showing balance sheets, investments, etc.


src/components/DateRangeInput.tsx
```diff
@@ -37,9 +29,11 @@ export default function DateRangeInput({
     const yearDiff = value as number;
     const key = now.year - yearDiff;
     shortcuts[key] = {
-      text: `End of ${now.year - yearDiff}`,
+      text: `Year ${now.year - yearDiff}`,
```
argaen, line 32: the shortcuts are now yearly (useful for seeing a summary of whole previous years)

PR #771, opened 2024-03-23
Cashflow interval

argaen:
Also changed the interval duration, when you selected 6 months it would show 7 months. Now I'm deducting 5 months so it will always show 6 bars/data points. On the last day of the month it will show the full 6 months.

PS: discovered a bug where if there is transfers between asset accounts that are different commodity in the cash flow, it is wrong. See:

Screenshot 2024-03-23 at 10 41 05 AM

we didn't transfer 17 euros, 17 is the amount of stocks bought. In `useCashFlow` we should convert everything to main currency :/


src/components/DateRangeInput.tsx
```diff
@@ -13,10 +13,17 @@ export default function DateRangeInput(): JSX.Element {
...
   const now = DateTime.now();
   const shortcuts: { [key: string]: { text: string, period: { start: Date, end: Date } } } = {
-    t: {
+    t3: {
+      text: 'Last 3 months',
+      period: {
+        start: now.minus({ months: 2 }).startOf('month').toJSDate(),
+        end: now.toJSDate(),
+      },
+    },
```
argaen, line 22: last 3 months shortcut is useful for taxes and so on

src/hooks/api/useCashFlow.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Returns an array containing the account guid, the total for that account, the type and
+ * the name.
+ */
+export function useCashFlow(
+  account: string,
+  selectedInterval?: Interval,
+): UseQueryResult<{ guid: string, total: number, type: string, name: string }[]> {
+  const { data: defaultInterval } = useInterval();
+  const interval = selectedInterval || defaultInterval;
```
argaen, line 19: nothing changed except this where we can pass a custom interval or use `useInterval` by default

PR #773, opened 2024-03-25
Cash flow in currency

argaen:
Fixed an issue with cash flow calculation where currencies of the accounts were not being take into account. In this version, we now retrieve each split in a transaction related to the account we are generating the cash flow report on and convert the quantities to the currency of the account.

To do that we use the following formula: `(main_split.quantity / main_split.value) * other_split.value`. The beauty of this is that it works for any kind of transaction regardless on the amount of splits it has or the currencies it has.

This is how the "Broker cash" account of the demo account looks like now:

Screenshot 2024-03-25 at 12 47 17 PM

vs before

Screenshot 2024-03-25 at 12 47 55 PM

> 35 was the number of stocks, not the euros amount

### DIFF
--- a/src/hooks/api/useCashFlow.ts
+++ b/src/hooks/api/useCashFlow.ts
@@ -14,10 +14,11 @@ export type CashFlowRow = {
 };
 
 /**
- * Retrieve the cashflow for a given account in the specified date.
+ * Retrieve the cashflow for a given account in the specified period.
  *
- * Returns an array containing the account guid, the total for that account, the type and
- * the name.
+ * The query works for transactions in multiple splits and in different currencies.
+ * We do that by calculating the exchange rate in the transaction and applying it to
+ * the splits for the other accounts.
  */
 export function useCashFlow(
   account: string,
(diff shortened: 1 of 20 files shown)


src/hooks/api/useCashFlow.ts
```diff
@@ -4,44 +4,71 @@ import type { UseQueryResult } from '@tanstack/react-query';
...
+        WHERE tx.guid IN (
+          SELECT DISTINCT tx_guid
+          FROM splits
+          WHERE account_guid = '${account}'
+        )
+        AND accounts.guid != '${account}'
+        AND tx.post_date >= '${((interval as Interval).start as DateTime).toSQLDate()}'
+        AND tx.post_date <= '${((interval as Interval).end as DateTime).toSQLDate()}'
+
+        GROUP BY splits.account_guid, accounts.name, accounts.account_type, mnemonic
```
argaen, line 60: well this query took quite a while to put together... basically the strategy it follows is:

1. grab each split and tie it to each account except for the account we are passing
2. grab the main split of each transaction
3. calculate the total for each split as `main_split.quantity / main_split.value * split.value`. This gives the total for that split **in the main account's currency**. This is what we want for the cash flow report.
4. hardcode the commodity to the commodity of the account we are generating the cash flow for.

src/hooks/api/useCashFlow.ts
```diff
@@ -4,44 +4,71 @@ import type { UseQueryResult } from '@tanstack/react-query';
...
+        JOIN splits ON splits.tx_guid = tx.guid
+        JOIN accounts ON splits.account_guid = accounts.guid
+        JOIN splits AS main_split ON main_split.tx_guid = tx.guid AND main_split.account_guid = '${account}'
+
+        WHERE tx.guid IN (
+          SELECT DISTINCT tx_guid
+          FROM splits
+          WHERE account_guid = '${account}'
+        )
+        AND accounts.guid != '${account}'
```
argaen, line 56: This avoids having an extra row for the account we generate the report for. We could add this as a double check to check that the quantities match but no use case for now.

PR #775, opened 2024-03-25
Add missing optiong to keep previous query data

argaen:
To prevent UI from flickering when user selects new account/interval/etc it's better to have `keepPreviousData` as default and disable it on specific queries when needed.

### DIFF
--- a/src/components/charts/MonthlyTotalHistogram.tsx
+++ b/src/components/charts/MonthlyTotalHistogram.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { DateTime, Interval } from 'luxon';
 import type { ChartDataset } from 'chart.js';
 
 import Bar from '@/components/charts/Bar';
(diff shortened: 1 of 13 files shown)


src/components/charts/MonthlyTotalHistogram.tsx
```diff
@@ -35,7 +30,7 @@ export default function MonthlyTotalHistogram({
   if (accounts && monthlyTotals) {
     guids.forEach(guid => {
       const data = dates.map((_, i) => (
-        monthlyTotals[i][guid]?.toNumber() || 0
+        monthlyTotals[i]?.[guid]?.toNumber() || 0
```
argaen, line 33: this is needed now because eventual consistency between interval and monthlyTotals

src/helpers/accountsTotalAggregations.ts
```diff
@@ -40,7 +40,7 @@ function aggregateWorth(
   dates.forEach((_, i) => {
     const zero = new Money(0, current.commodity?.mnemonic);
     const previousMonth = aggregatedTotals[i - 1]?.[current.guid] || zero;
-    const currentMonth = monthlyTotals[i][current.guid] || zero;
+    const currentMonth = monthlyTotals[i]?.[current.guid] || zero;
```
argaen, line 43: this is needed now because eventual consistency between interval and monthlyTotals

PR #781, opened 2024-03-26
Income statement and balance sheet

argaen:
Splitting accountsTotals to balance sheet and income statement. In the end they calculate the same with couple of differences:

- balance sheet only takes an end date while income statement takes a period.
- income statement doesn't convert prices because for now we only let income/expense accounts to be in main currency

We still have accountsTotals functionality for cases we just need an account "total" which for some cases it is the balance sheet while for others is the income statement.

With this change, all the widgets (except the investments section) now make use of the interval defined by the user. All the numbers match **except** the networth histogram and the total change difference. The reason is because in the difference we calculate the period difference while in the networth histogram, we display the networth at the end of the month for each data point.

For example, if the user selects 1st May to 1st July, we display the networth at the end of May, June and 1st July, while the difference is 1st July - 1st May net worth.

I haven't thought yet how to solve this, couple of options:

- In the tooltip we display the date with the day that represents that data point
- The first point in the net worth histogram displays net worth on the start instead of automatically selecting the end of the month.


src/book/Money.ts
```diff
@@ -9,6 +9,7 @@ import type { Commodity } from './entities';
 
 export default class Money {
   private _raw: djs.Dinero<number>;
+  readable: string;
```
argaen, line 12: so we can read the money object when debugging/printing...

PR #783, opened 2024-03-26
Return currency on missing price

argaen:


src/app/dashboard/layout.tsx
```diff
@@ -32,15 +32,13 @@ export default function DashboardLayout({
...
-
-              return error.show();
+          FallbackComponent={ErrorComponent}
+          onError={(error) => {
+            if (!(error instanceof MaffinError)) {
+              error = new MaffinError(error.message, 'UNKNOWN');
             }
-          }
+            (error as MaffinError).show();
+          }}
```
argaen, line 41: So when there is an error triggered, navigation stops working. This is because the state of ErrorBoundary still has an error and thus, keeps rendering the fallback component. I... tried to fix this but haven't managed to :/

PR #786, opened 2024-03-27
Fix net worth intervals

argaen:
When changing partial intervals, the net worth graph was being computed incorrectly. We now display it as expected.

One thing to note as example:

when you select 1st March to any date, the net worth graph will display last day of March as the first point. This can be confusing at start in the assets detail page because it shows the growth during that period which missaligns with the quantities shown in the networth graph (if you make the difference). This is on purpose, because as mentioned, the first point is displayed at the end of the month to give consistency on the dates we display the net worth calculation (always at the last day of the month except for the end of the selected interval).

I tried to display also net worth on the start of the interval but it is messy as it will show two points in the same month. We can always revisit in the future


src/components/charts/MonthlyTotalHistogram.tsx
```diff
@@ -5,8 +5,8 @@ import Bar from '@/components/charts/Bar';
 import type { Account } from '@/book/entities';
 import { useAccounts, useAccountsMonthlyTotal, useMainCurrency } from '@/hooks/api';
 import { moneyToString } from '@/helpers/number';
-import monthlyDates from '@/helpers/monthlyDates';
 import { useInterval } from '@/hooks/state';
+import { intervalToDates } from '@/helpers/dates';
```
argaen, line 9: removing monthlyDates in favor of intervalToDates which is more data centric

src/components/charts/NetWorthHistogram.tsx
```diff
@@ -97,94 +97,97 @@ export default function NetWorthHistogram({
...
+                  usePointStyle: true,
+                  pointStyle: 'circle',
+                  boxHeight: 8,
+                  boxWidth: 8,
+                },
+              },
+              tooltip: {
+                backgroundColor: '#323b44',
+                callbacks: {
+                  title: (tooltipItem) => labels[tooltipItem[0].dataIndex].toFormat('DD'),
```
argaen, line 177: we now display the day the data is computed to the user in this chart so they it gives clarity

src/hooks/api/useMonthlyWorth.ts
```diff
@@ -33,31 +33,29 @@ export function useMonthlyWorth(
...
 
+      const dates = [interval.start as DateTime, ...intervalToDates(interval as Interval)];
       return aggregated.map((d, i) => aggregateChildrenTotals(
         ['type_asset', 'type_liability'],
         accounts as Account[],
         prices as PriceDBMap,
         dates[i],
         d,
-      ));
+      )).slice(1);
```
argaen, line 58: we now compute the previous total until the start of the interval and then add it to the monthly worth. Once added, we can remove the first array as otherwise it creates a series with 1 extra data point

src/hooks/api/useMonthlyWorth.ts
```diff
@@ -77,10 +75,7 @@ export function useMonthlyWorth(
     queryFn: fetcher(
       () => getMonthlyTotals(
         accounts as Account[],
-        Interval.fromDateTimes(
-          ((interval as Interval).start as DateTime).plus({ month: 1 }),
-          (interval as Interval).end as DateTime,
-        ),
+        interval,
```
argaen, line 78: check comment above which simplifies dealing with interval

PR #793, opened 2024-03-28
Investment chart interval

argaen:
Also as extra:

- useInvestment is a select on useInvestments, same as it was for useAccount/s.
- useInvestments now selects the date to process splits based on the user selected interval
- Moved useAccountsMonthlyTotal to its own file and called it useMonthlyTotals


src/components/pages/account/InvestmentChart.tsx
```diff
@@ -128,6 +123,13 @@ export default function InvestmentChart({
           tooltip: {
             backgroundColor: '#323b44',
             callbacks: {
+              title: (tooltipItems) => {
+                if (tooltipItems[0].datasetIndex === 0) {
+                  return dates[tooltipItems[0].dataIndex].toFormat('DD');
```
argaen, line 128: this is kind of a hack. We draw the bar in the 1st of the month to align it with the tick but in reality we've calculated at the last day of the month.

src/components/pages/account/InvestmentChart.tsx
```diff
@@ -49,14 +49,12 @@ export default function InvestmentChart({
...
-    ? intervalToDates(interval).map(d => d.startOf('month'))
-    : interval.splitBy({ day: 1 }).map(d => (d.start as DateTime));
-  monthly.forEach(x => {
+  const dates = intervalToDates(interval);
+
+  dates.forEach(x => {
     investment.processSplits(x);
     numStocksData.push({
-      x: x.toMillis(),
+      x: x.startOf('month').toMillis(),
```
argaen, line 57: hack to align bars at the tick of the month

PR #795, opened 2024-03-28
Fix query invalidation for txs

argaen:
When I refactored the accounts total query to use balance-sheet and income-statement forgot to update the keys so numbers where updated on new transactions

### DIFF
--- a/src/book/entities/Transaction.ts
+++ b/src/book/entities/Transaction.ts
@@ -121,18 +121,8 @@ export async function updateCache(
     queryKey: [...Transaction.CACHE_KEY, { name: 'start' }],
   });
 
-  entity.splits.forEach(split => {
-    queryClient.invalidateQueries({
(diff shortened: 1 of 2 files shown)


src/book/entities/Transaction.ts
```diff
@@ -121,18 +121,8 @@ export async function updateCache(
...
-    });
-  });
-
-  queryClient.invalidateQueries({
-    queryKey: [...Split.CACHE_KEY, { aggregation: 'total' }],
-  });
-
   queryClient.invalidateQueries({
-    queryKey: [...Split.CACHE_KEY, { aggregation: 'monthly-total' }],
+    queryKey: [...Split.CACHE_KEY],
```
argaen, line 125: Given that now we have data bounded by intervals and better aggregation, we can just purge all splits as retrieving this data is not so costly.

PR #797, opened 2024-03-28
Fix multiple split

argaen:
Well, the SQL became more complicated but it covers the use cases

### DIFF
--- a/src/hooks/api/useCashFlow.ts
+++ b/src/hooks/api/useCashFlow.ts
@@ -44,12 +44,12 @@ export function useCashFlow(
               JOIN accounts AS main_account ON main_account.guid = '${account}'
               WHERE commodities.guid = main_account.commodity_guid
             ) AS mnemonic,
-            SUM(((cast(main_split.quantity_num as REAL) / main_split.quantity_denom) / (cast(main_split.value_num as REAL) / main_split.value_denom)) * (cast(splits.value_num as REAL) / splits.value_denom)) as total
+            SUM(((cast(account_split.quantity_num as REAL) / account_split.quantity_denom) / (cast(account_split.value_num as REAL) / account_split.value_denom)) * (cast(splits.value_num as REAL) / splits.value_denom)) as total
 
           FROM transactions AS tx
           JOIN splits ON splits.tx_guid = tx.guid
           JOIN accounts ON splits.account_guid = accounts.guid
-          JOIN splits AS main_split ON main_split.tx_guid = tx.guid AND main_split.account_guid = '${account}'
+          JOIN splits AS account_split ON account_split.tx_guid = tx.guid AND account_split.account_guid = '${account}'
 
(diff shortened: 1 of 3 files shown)


src/hooks/api/useCashFlow.ts
```diff
@@ -44,12 +44,25 @@ export function useCashFlow(
...
+                  END
+              END
+            ) AS total
 
           FROM transactions AS tx
           JOIN splits ON splits.tx_guid = tx.guid
           JOIN accounts ON splits.account_guid = accounts.guid
-          JOIN splits AS main_split ON main_split.tx_guid = tx.guid AND main_split.account_guid = '${account}'
+          JOIN splits AS account_split ON account_split.tx_guid = tx.guid AND account_split.account_guid = '${account}'
+          JOIN splits AS credit_split ON credit_split.tx_guid = tx.guid AND credit_split.value_num < 0
```
argaen, line 65: we only support having 1 credit split while there can be multiple debit splits. We use the credit_split to calculate partials for the cashflow

src/hooks/api/useCashFlow.ts
```diff
@@ -44,12 +44,25 @@ export function useCashFlow(
               JOIN accounts AS main_account ON main_account.guid = '${account}'
               WHERE commodities.guid = main_account.commodity_guid
             ) AS mnemonic,
-            SUM(((cast(main_split.quantity_num as REAL) / main_split.quantity_denom) / (cast(main_split.value_num as REAL) / main_split.value_denom)) * (cast(splits.value_num as REAL) / splits.value_denom)) as total
+            SUM(
+              CASE
+                WHEN splits.guid = credit_split.guid THEN
+                  -1 * cast(account_split.quantity_num AS REAL) / account_split.quantity_denom
```
argaen, line 50: for the case where the split we are calculating for is the credit_split, we instead pick the quantity from the account_split and reference it to the credit account. An example:

```
Assets:Bank -100
Liabilities:L1 75
Expenses:E2 25
```

When we compute cashflow for L1 (cashflow is available for both assets and liabilities) we want it to be:

`Assets:Bank -75`

src/hooks/api/useCashFlow.ts
```diff
@@ -44,12 +44,25 @@ export function useCashFlow(
...
+                ELSE
+                  CASE
+                    WHEN account_split.guid = credit_split.guid THEN
+                      (cast(account_split.quantity_num AS REAL) / account_split.quantity_denom) / (cast(account_split.value_num AS REAL) / account_split.value_denom) * (cast(splits.value_num AS REAL) / splits.value_denom)
```
argaen, line 54: When the account_split is the same as the credit_split then we pick the splits value and convert it to the account_split currency. This helps for when other splits and the account_split are in different currencies so we display the cashflow in the currency of the account we are calculating for

PR #800, opened 2024-03-29
Set price on processing splits

argaen:
Set the price in the investment account to be on the date we are processing the splits. This is so we show the right value at the selected date.

### DIFF
--- a/src/components/pages/account/InvestmentInfo.tsx
+++ b/src/components/pages/account/InvestmentInfo.tsx
@@ -1,14 +1,12 @@
 import React from 'react';
 import classNames from 'classnames';
 import dynamic from 'next/dynamic';
-import { DateTime } from 'luxon';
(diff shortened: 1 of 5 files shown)


src/components/pages/account/InvestmentInfo.tsx
```diff
@@ -21,20 +19,12 @@ export type InvestmentInfoProps = {
...
-  if (!investment || !prices) {
+  if (!investment) {
     return <Loading />;
   }
 
-  prices = prices || [];
-  const latestPrice = prices.getInvestmentPrice(
-    investment.account.commodity.mnemonic,
-    interval.end as DateTime,
-  );
```
argaen, line 36: We already have the current price inside the InvestmentAccount thanks to this change so removing redundant data and using the info inside `quoteInfo`

PR #809, opened 2024-04-02
Add split and dividend actions

argaen:
Add "split" and "dividend" actions in the transaction dropdown for investments.

Also as a bonus, fixed the issue where adding new transactions to an investment doesn't refresh their numbers


src/components/buttons/AddTxDropdown.tsx
```diff
@@ -50,6 +55,87 @@ export default function AddTxDropdown({
...
+                          Split.create({
+                            fk_account: account,
+                            valueNum: 0,
+                            valueDenom: 1,
+                            quantityNum: 0,
+                            quantityDenom: 1,
+                          }),
+                          Split.create({
+                            fk_account: {
+                              type: 'INCOME',
```
argaen, line 90: this makes sure only INCOME accounts are selected

src/components/forms/transaction/MainSplit.tsx
```diff
@@ -52,11 +54,11 @@ export default function MainSplit({
   }, [txCurrency, date, form.formState.isDirty, account, prices]);
 
   React.useEffect(() => {
-    if (form.formState.isDirty) {
+    if (form.formState.isDirty && splits.length > 1) {
```
argaen, line 57: This is so we support split events for investments. I don't think there is another case where splits length is only 1 so using this condition for now.

src/components/forms/transaction/SplitField.tsx
```diff
@@ -80,8 +81,9 @@ export default function SplitField({
                 id={`splits.${index}.account`}
                 isClearable={false}
                 isDisabled={disabled}
-                placeholder="<account>"
+                placeholder={account?.type ? `<Choose a ${account.type} account>` : '<Choose an account>'}
                 ignorePlaceholders
+                onlyTypes={account?.type ? getAllowedSubAccounts(account.type) : undefined}
```
argaen, line 86: new condition for AccountSelector that allows to filter by type.

src/hooks/api/useInvestments.ts
```diff
@@ -36,20 +38,28 @@ export function useInvestments<TData = InvestmentAccount[]>(
 ): UseQueryResult<TData> {
   const { data: interval } = useInterval();
 
-  const { data: accounts } = useAccounts();
+  const { data: accounts, dataUpdatedAt: accountsUpdatedAt } = useAccounts();
+  const { data: splits, dataUpdatedAt: splitsUpdatedAt } = useSplits({ type: 'INVESTMENT' });
```
argaen, line 42: bringing splits back into the query so values are refreshed when splits are updated

PR #811, opened 2024-04-02
Support for roboadvisors

argaen:
Some changes to support robo advisor investments:

- Robo advisor investments are accounts categorised as "INVESTMENT"
- They use a currency (before it was not possible)

Screenshot 2024-04-02 at 5 16 37 PM

### DIFF
--- a/src/book/entities/Account.ts
+++ b/src/book/entities/Account.ts
@@ -13,7 +13,6 @@ import {
 import type { QueryClient } from '@tanstack/react-query';
 
 import {
-  isInvestment,
   getAllowedSubAccounts,
   ASSET_ACCOUNTS,
   LIABILITY_ACCOUNTS,
@@ -90,7 +89,6 @@ export default class Account extends BaseEntity {
 
   @ManyToOne('Commodity', { eager: true, cascade: true })
   @JoinColumn({ name: 'commodity_guid' })
-  @CheckInvestmentCommodity()
   @CheckIECommodity()
   @v.ValidateIf(o => o.type !== 'ROOT')
   @v.IsNotEmpty({ message: 'commodity is required' })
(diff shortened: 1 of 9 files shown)


src/book/entities/Account.ts
```diff
@@ -185,34 +183,6 @@ function CheckAccountType(validationOptions?: v.ValidationOptions) {
...
-          return true;
-        },
-
-        defaultMessage() {
-          return 'investment accounts cant have a currency as their commodity';
-        },
-      },
-    });
-  };
-}
```
argaen, line 214: not anymore, robo advisors can be in the currency. You loose information because theoretically you should track all the share movements but that is too much in some cases

src/book/models/InvestmentAccount.ts
```diff
@@ -42,9 +42,6 @@ export default class InvestmentAccount {
       this.account.commodity.mnemonic,
       DateTime.now(),
     );
-    if (price.guid === 'missing_price') {
-      throw new Error(`No price found for ${this.account.commodity.mnemonic}`);
-    }
```
argaen, line 47: this is because when it's in currency, we will returning missing_price with value 1.

src/components/pages/account/InvestmentChart.tsx
```diff
@@ -50,61 +52,161 @@ export default function InvestmentChart({
...
+        investment.account.commodity.mnemonic,
+        s.transaction.date,
+      );
+
+      if (price) {
+        valueData.push({
+          x: s.transaction.date.toMillis(),
+          y: investment.quantity.toNumber() * price.value,
+        });
+      }
```
argaen, line 76: so basically we are changing the x axis data points. Now we show a point every time there is a buy/sell/split movement plus the last day of each month (if it doesn't exist already).

src/components/pages/account/InvestmentChart.tsx
```diff
@@ -50,61 +52,161 @@ export default function InvestmentChart({
...
-      valueData.push({
+  dates.forEach(x => {
+    const alreadyExists = numStocksData.find(d => d.x === x.toMillis());
+
+    if (!alreadyExists) {
+      investment.processSplits(x);
+      numStocksData.push({
         x: x.toMillis(),
-        y: investment.quantity.toNumber() * price.value,
+        y: !isCurrencyInvestment ? investment.quantity.toNumber() : investment.cost.toNumber(),
```
argaen, line 86: when it is in currency, we want to display the cost as "quantity" will show the value (because there is no alternative commodity)

src/components/pages/account/InvestmentChart.tsx
```diff
@@ -50,61 +52,161 @@ export default function InvestmentChart({
...
+      },
+      ticks: {
+        color: 'rgba(22, 163, 74)',
+        maxTicksLimit: 10,
+        callback: (value) => moneyToString(value as number, currency),
+      },
+    },
+  };
+
+  if (!isCurrencyInvestment) {
```
argaen, line 176: we only show the price if it's a non currency commodity

PR #816, opened 2024-04-03
Investment placeholder

argaen:
Removed the generic "investments" page. Now the user can have investment placeholder pages and those will show the investments page filtered by the investments in the category. This gives more control to the users to organise their investments in different sections and get the reports based on those.

Also as bonus, added emojis to the demo.


src/book/entities/Account.ts
```diff
@@ -243,10 +248,27 @@ function CheckPlaceholder(validationOptions?: v.ValidationOptions) {
             }
           }
 
+          if (account.childrenIds?.length > 0 && !account.placeholder) {
+            return false;
+          }
+
+          if (account.parent?.type === 'INVESTMENT' && account.placeholder) {
+            return false;
+          }
```
argaen, line 257: We don't allow multiple levels of nesting placeholder INVESTMENT accounts. The very top level should be an 'ASSET'

src/book/models/InvestmentAccount.ts
```diff
@@ -223,7 +223,7 @@ export default class InvestmentAccount {
 
   static isSplit(split: Split): boolean {
     const numSplits = split.transaction.splits.length;
-    return numSplits === 1 && split.value === 0 && split.quantity > 0;
+    return numSplits === 1 && split.value === 0 && split.quantity !== 0;
```
argaen, line 226: With previous code we only allowed split events but not reverse split events (i.e. number of stocks decreases). This is necessary for updating the Robo advisor prices with losses

src/components/forms/transaction/SplitField.tsx
```diff
@@ -144,7 +145,7 @@ export default function SplitField({
             step="0.001"
             disabled={disabled}
             onChange={(e) => {
-              if (account.commodity.guid === txCurrency.guid) {
+              if (!showValueField && splits.length > 1) {
```
argaen, line 148: When it is an update, we display the single split of a split event as a SplitField and not a MainSplit. This is to avoid the quantity being updated

src/hooks/api/useInvestments.ts
```diff
@@ -53,7 +56,7 @@ export function useInvestments<TData = InvestmentAccount[]>(
     queryKey,
     queryFn: fetcher(
       () => getInvestments(
-        (accounts as Account[]).filter(a => a.type === 'INVESTMENT'),
+        (accounts as Account[]).filter(a => a.type === 'INVESTMENT' && !a.placeholder),
```
argaen, line 59: We still calculate for ALL investments at once for now. The way we filter is by using the `select` function

PR #817, opened 2024-04-03
Delete unused commodities

argaen:
Let user delete unused commodities

### DIFF
--- a/src/book/entities/Account.ts
+++ b/src/book/entities/Account.ts
@@ -134,11 +134,12 @@ export default class Account extends BaseEntity {
   }
 
   async remove(options?: SaveOptions): Promise<this> {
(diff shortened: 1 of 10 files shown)


src/book/entities/Account.ts
```diff
@@ -134,11 +134,12 @@ export default class Account extends BaseEntity {
   }
 
   async remove(options?: SaveOptions): Promise<this> {
+    const account = await super.remove(options);
+
     if (this.queryClient) {
       updateCache({ queryClient: this.queryClient });
     }
 
-    const account = await super.remove(options);
```
argaen, line 141: I'm not sure why it was in this order but this creates an issue where the cache is updated before it is deleted.

PR #821, opened 2024-04-04
Use cost for total invested

argaen:
Using total bought is wrong as it counts sold titles too thus showing wrong number.

Also, increased the resolution of `format` call so it uses 4 decimals and then rounds instead of using 2 decimals. This reduces decimal error.

### DIFF
--- a/src/book/Money.ts
+++ b/src/book/Money.ts
@@ -53,10 +53,10 @@ export default class Money {
    * Represents the money object as a string with the currency standardizing
    * the scale to 2 as a default. If the scale is passed then it uses that.
    */
-  toString(scale = 2): string {
+  toString(scale = 4): string {
     return djs.toDecimal(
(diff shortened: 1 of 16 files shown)


src/book/Money.ts
```diff
@@ -66,10 +66,10 @@ export default class Money {
    *
    * The result is a localized string with the currency as a symbol
    */
-  format(scale = 2): string {
+  format(scale = 4): string {
     return djs.toDecimal(
       djs.transformScale(this._raw, scale),
-      ({ value, currency }) => moneyToString(Number(value), currency.code),
+      ({ value, currency }) => moneyToString(toFixed(Number(value)), currency.code),
```
argaen, line 72: We round to two decimals (from 4). This is for cases where something is 1.9999, should become 2

PR #827, opened 2024-04-05
Improve commodities sync

argaen:
Improve some commodities behaviors:

- Stocker was failing the whole request when 1 of the tickers passed failed which made all commodities to be un-synced. Now, if one of them fails, we just ignore it.
- We display a warning for commodities that don't have any prices
Screenshot 2024-04-05 at 12 15 32 PM

- Whenever a new currency or stock/mutual is added, we re-sync all prices.

There is a new behavior I found that is a bit weird and yet unsolved:

- If you add a new STOCK let's say TSLA which has a base currency of USD but you don't have that currency, it will show the warning. This is solved whenever the user adds USD (and both prices will be synced) but the error message shown to the user is not trivial.

### DIFF
--- a/amplify/backend/function/stockerlambda/ts/app.ts
+++ b/amplify/backend/function/stockerlambda/ts/app.ts
@@ -43,7 +43,11 @@ app.get('/api/prices', async (req, res) => {
   const result: { [ticker: string]: Price } = {};
 
   async function callAndSave(ticker: string) {
-    result[ticker] = await yahoo.getPrice(ticker);
+    try {
+      result[ticker] = await yahoo.getPrice(ticker);
+    } catch (e) {
+      console.warn(`A price retrieval failed: ${e}`);
+    }
   }
 
   const promises: Promise<void>[] = [];
--- a/src/app/dashboard/commodities/page.tsx
+++ b/src/app/dashboard/commodities/page.tsx
@@ -2,21 +2,18 @@
 
 import React from 'react';
 import {
-  BiLineChart,
-  BiMoney,
-  BiSolidGhost,
   BiPlusCircle,
 } from 'react-icons/bi';
 
 import { useCommodities } from '@/hooks/api';
-import Link from 'next/link';
 import Loading from '@/components/Loading';
 import FormButton from '@/components/buttons/FormButton';
 import CommodityForm from '@/components/forms/commodity/CommodityForm';
+import { insertTodayPrices } from '@/lib/Stocker';
+import CommodityCard from '@/components/CommodityCard';
 
 export default function CommoditiesPage(): JSX.Element {
-  let { data: commodities } = useCommodities();
-  const { isLoading } = useCommodities();
+  const { data: commodities = [], isLoading } = useCommodities();
 
   if (isLoading) {
     return (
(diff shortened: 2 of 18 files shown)


amplify/backend/function/stockerlambda/ts/app.ts
```diff
@@ -43,7 +43,11 @@ app.get('/api/prices', async (req, res) => {
   const result: { [ticker: string]: Price } = {};
 
   async function callAndSave(ticker: string) {
-    result[ticker] = await yahoo.getPrice(ticker);
+    try {
+      result[ticker] = await yahoo.getPrice(ticker);
+    } catch (e) {
+      console.warn(`A price retrieval failed: ${e}`);
```
argaen, line 49: ignore errors for failed retrievals, we still return a successful response.

src/app/dashboard/commodities/page.tsx
```diff
@@ -51,7 +46,7 @@ export default function CommoditiesPage(): JSX.Element {
               </>
             )}
           >
-            <CommodityForm />
+            <CommodityForm onSave={() => insertTodayPrices()} />
```
argaen, line 49: On creating a new commodity, re-sync all prices

src/app/dashboard/commodities/page.tsx
```diff
@@ -61,18 +56,7 @@ export default function CommoditiesPage(): JSX.Element {
...
-            href={`/dashboard/commodities/${commodity.guid}`}
-            className="card col-span-1 cursor-pointer hover:shadow-lg text-slate-400 dark:hover:text-white"
-          >
-            <div className="flex items-center">
-              <BiMoney className="mr-1" />
-              {commodity.mnemonic}
-            </div>
-          </Link>
-        ))}
+        {currencies.map(commodity => <CommodityCard key={commodity.guid} guid={commodity.guid} />)}
```
argaen, line 59: new CommodityCard component that contains all logic about the commodity card (including the warning, tooltip, etc)

src/components/forms/commodity/CommodityForm.tsx
```diff
@@ -112,6 +112,8 @@ async function onSubmit(
   if (action === 'add' || action === 'update') {
     await commodity.save();
   } else if (action === 'delete') {
+    Price.delete({ fk_commodity: commodity });
+    Price.delete({ fk_currency: commodity });
```
argaen, line 116: Delete prices related to that commodity.

Sadly this doesn't invalidate the cache because we've only implemented it for the instance.remove method. Maybe will improve in the future

src/hooks/useDataSource.tsx
```diff
@@ -176,13 +176,11 @@ async function createEmptyBook() {
...
 
 async function loadStockerPrices() {
-  if (IS_PAID_PLAN) {
-    try {
-      // We have to await because we use Price.upsert. Once we correct this,
-      // saving Price already updates the local cache so we don't need await
-      await insertTodayPrices();
-    } catch (e) {
-      console.warn(`Retrieving live prices failed ${e}`);
-    }
```
argaen, line 186: given we re-use the insertTodayPrices somewhere else, just move the logic of `IS_PAID_PLAN` within.

PR #835, opened 2024-04-07
Move transaction preloading for transaction form

argaen:
In previous version the caller of TransactionForm had to know which Transaction entity fields were needed. This is ugly because it moves knowledge out of the component and makes re-usability hard as each time you use the form, you have to repeat the same logic.

Changed it so:

- When action is add, you can still pass default values as needed
- When action is update or delete, you MUST pass a `guid` and not defaultValues. With this guid, the transaction form loads the transaction values needed to show it properly in the form.

### DIFF
--- a/src/components/forms/transaction/TransactionForm.tsx
+++ b/src/components/forms/transaction/TransactionForm.tsx
@@ -10,30 +10,67 @@ import {
   Transaction,
 } from '@/book/entities';
 import { Tooltip } from '@/components/tooltips';
+import Loading from '@/components/Loading';
+import { useTransaction } from '@/hooks/api';
 import SplitsField from './SplitsField';
 import type { FormValues } from './types';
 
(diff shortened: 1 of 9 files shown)


src/components/forms/transaction/TransactionForm.tsx
```diff
@@ -3,38 +3,76 @@ import { DateTime } from 'luxon';
...
+        // class to the form, then we have reference errors as when
+        // we update the form, it also updates the defaultValues
+        // which means formState.isDirty is not triggered properly
+        splits: data.splits.map(split => ({
+          ...split,
+          value: split.value,
+          quantity: split.quantity,
+          account: split.fk_account,
+        } as Split)),
+      } as FormValues
```
argaen, line 59: convert the Transaction entity to FormValues so it can be parsed by the form directly.

PR #837, opened 2024-04-08
Minor improvements to liability page

### LINKED ISSUE #834: Liabilities dashboard is not intuitive
The logic is correct for liabilities, however, the dahsboard displayed makes one think that my liability amount is going up, even when I am making repayments. 

Changing to absolute numbers would be more user-friendly to tick the right boxes in a user's brain.
Screenshot 2024-04-07 at 7 24 57 PM

argaen:
I've changed the networth histogram to be a bit more configurable so now it can:

- Show assets, liabilities and net worth
- Show only asset account with its own configuration
- Show only liability account with its own configuration

Screenshot 2024-04-08 at 12 14 56 PM

I've decided to not make the values for liabilities absolute as we are talking about debt. Saying that at some point in time the amount is -934 is correct. Would be confusing to say its 934 as the users may think they are in positive.

Fixes #834

### DIFF
--- a/src/components/charts/NetWorthHistogram.tsx
+++ b/src/components/charts/NetWorthHistogram.tsx
@@ -20,24 +20,41 @@ C.register(
   LineController,
 );
 
-export type NetWorthHistogramProps = {
+export type NetWorthHistogramProps =
+| {
+  title?: string,
   assetsGuid: string,
-  assetsLabel?: string,
-  hideAssets?: boolean,
+  assetsConfig?: Partial<ChartDataset>,
+  liabilitiesGuid?: never,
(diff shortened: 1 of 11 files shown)


src/components/charts/NetWorthHistogram.tsx
```diff
@@ -20,24 +20,41 @@ C.register(
   LineController,
 );
 
-export type NetWorthHistogramProps = {
+export type NetWorthHistogramProps =
```
argaen, line 23: users can now pass either only assetsGuid or liabilitiesGuid or both at the same time.

PR #838, opened 2024-04-08
Fix account selection in account form

### LINKED ISSUE #832: Cannot reselect other account types when creating transactions
If you accidentally select a wrong account while entering transactions, Maffin doesn't let you reselect a new account, even after you close the transaction window and click [Add] again. 
Screenshot 2024-04-07 at 10 57 18 PM

argaen:
An old bug... passing an instance of a Split object makes the defaultValue reference to be updated when changing the value. Passing now a {} instead of a split instance in the add transaction second split to avoid this and then using `formState.defaultValues` to see if we have set a default for the account type.

Fixes #832

### DIFF
--- a/src/components/buttons/AddTxDropdown.tsx
+++ b/src/components/buttons/AddTxDropdown.tsx
@@ -48,7 +48,10 @@ export default function AddTxDropdown({
                 {
                   date: (latestDate || DateTime.now()).toISODate() as string,
                   description: '',
-                  splits: [Split.create({ fk_account: account }), new Split()],
+                  splits: [
+                    Split.create({ fk_account: account }),
+                    {} as Split,
+                  ],
                   fk_currency: account.commodity,
                 }
               }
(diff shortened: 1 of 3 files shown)


src/components/buttons/AddTxDropdown.tsx
```diff
@@ -48,7 +48,10 @@ export default function AddTxDropdown({
                 {
                   date: (latestDate || DateTime.now()).toISODate() as string,
                   description: '',
-                  splits: [Split.create({ fk_account: account }), new Split()],
+                  splits: [
+                    Split.create({ fk_account: account }),
+                    {},
```
argaen, line 53: passing a split object makes it that when we select an account for this split, the `defaultValues` object is also updated. Setting it to a raw object so this doesn't happen.

src/components/forms/transaction/SplitField.tsx
```diff
@@ -69,6 +69,8 @@ export default function SplitField({
     form.trigger('splits');
   }, [value, exchangeRate, form, action, index]);
 
+  const defaultAccount = form.formState.defaultValues?.splits?.[index]?.fk_account as Account;
```
argaen, line 72: use the `formState.defaultValues` instead of account watcher to narrow user options.

PR #839, opened 2024-04-08
Tx form account selection

### LINKED ISSUE #833: Persistent Dinero.js error code on investment account
Trying to track my crypto wallet overall value change the same way as robo-advisors, but even when the commodity is connected properly, a dinero.js error pops up when I tried to create a transaction.

The transaction involves receiving part of my bonus (as salary) in crypto which is converted from USD -> SGD.

After this error code appears, nothing else can be done to the investment account, not even deleting it.
Screenshot 2024-04-07 at 4 42 37 PM

argaen:
When we added support for RoboAdvisors we made the change that investment accounts can be assigned a currency as commodity. This broke the `txCurrency` selection in the Transaction form as in there we assumed that investment accounts would always have a non currency.

This fixes this by checking in the account of the split is of INVESTMENT type instead of checking the commodity to decide the txCurrency

> We always want the transactions for an investment to be in the currency of the investment itself so the computations done in InvestmentAccount are correct. Otherwise we start mixing currencies and it will throw errors.

Fixes #833

### DIFF
--- a/src/components/forms/transaction/SplitField.tsx
+++ b/src/components/forms/transaction/SplitField.tsx
@@ -69,6 +69,8 @@ export default function SplitField({
     form.trigger('splits');
   }, [value, exchangeRate, form, action, index]);
 
+  const defaultAccount = form.formState.defaultValues?.splits?.[index]?.fk_account as Account;
+
   return (
     <fieldset className="grid grid-cols-13" key={`splits.${index}`}>
       <div className={`col-span-${showValueField ? 7 : 9}`}>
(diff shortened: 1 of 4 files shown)


src/components/forms/transaction/SplitField.tsx
```diff
@@ -94,17 +94,22 @@ export default function SplitField({
...
-                        fk_commodity: {
-                          guid: (nonCurrencySplit.fk_account as Account).commodity.guid,
-                        },
-                      });
-                      form.setValue('fk_currency', price.currency);
+                    if (investmentSplit) {
+                      const investmentAccount = investmentSplit.fk_account as Account;
+                      let currency = investmentAccount.commodity;
+
+                      if (investmentAccount.commodity.namespace !== 'CURRENCY') {
```
argaen, line 105: if the investment has a non currency commodity, then retrieve the currency associated to it.

src/components/forms/transaction/SplitField.tsx
```diff
@@ -94,17 +94,25 @@ export default function SplitField({
                   const splits = form.getValues('splits');
 
                   if (a) {
-                    const nonCurrencySplit = splits.find(
-                      split => (split.fk_account as Account).commodity.namespace !== 'CURRENCY',
+                    const investmentSplit = splits.find(
+                      split => (
+                        (split.fk_account as Account).type === 'INVESTMENT'
+                        || (split.fk_account as Account).commodity.namespace !== 'CURRENCY'
```
argaen, line 100: check for either non currency commodity or for the account being an investment.

PR #840, opened 2024-04-08
Improve yahoo price sync

### LINKED ISSUE #831: Currencies Commodities bug
1. Currency pairs happen between all currency commodities created on Prices pages, instead of just pairing with the main currency.
2. Latest prices are registering as 0 even though there were existing prices for a commodity before, specifically for lesser known currencies like IDR, THB, PHP. Seems like stronger currencies are fine.
3. Not sure if it's because of the pairs but charts are looking super weird.
Screenshot 2024-04-07 at 10 51 49 PM

argaen:
So apparently, for some currency rates, yahoo returns very bad resolution. Take https://finance.yahoo.com/quote/IDREUR=X?.tsrc=fin-srch for example which always returns 0.0001 which is definitely wrong.

The change I've introduced is see if there is some reliable currency in the ticker ('SGD', 'USD' and 'EUR') for now and if there is, use that one as a base

I.E IDREUR=X returns crap but EURIDR=X works as expected. So we apply this behavior too.

This is why in #831 IDR is always 10000. After this change, it shows correct prices:

Screenshot 2024-04-08 at 4 51 18 PM

Also, as an improvement for this, increased the amount of decimals that can be shown in the prices chart and prices table

Fixes #831

### DIFF
--- a/amplify/backend/function/stockerlambda/ts/yahoo.ts
+++ b/amplify/backend/function/stockerlambda/ts/yahoo.ts
@@ -8,10 +8,8 @@ export class YahooError extends HTTPError {}
 
 const HOST = 'https://query2.finance.yahoo.com';
 
-export async function getPrice(ticker: string, when?: number): Promise<Price> {
-  if (ticker === 'SGDCAD=X') {
-    ticker = 'SGDCAX=X';
-  }
+export async function getPrice(t: string, when?: number): Promise<Price> {
+  const { ticker, transform } = formatTicker(t);
   let url = `${HOST}/v8/finance/chart/${ticker}?interval=1d&includePrePost=false`;
 
   if (when) {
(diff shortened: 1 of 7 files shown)


amplify/backend/function/stockerlambda/ts/yahoo.ts
```diff
@@ -69,6 +67,30 @@ export async function getPrice(ticker: string, when?: number): Promise<Price> {
...
+  if (ticker === 'SGDCAD=X') {
+    return {
+      ticker: 'SGDCAX=X',
+      transform: (n: number) => n,
+    }
+  }
+
+  if (
+    /[EUR|USD|SGD]=X$/.test(ticker)
+    && !(/^[EUR|USD|SGD]/.test(ticker))
```
argaen, line 80: If ticker ends in EUR USD or SGD but doesn't start with any of those, then reverse the ticker so we get reliable rates from yahoo

src/helpers/number.ts
```diff
@@ -41,6 +41,7 @@ export function moneyToString(n: number, currency: string): string {
     return n.toLocaleString(navigator.language, {
       style: 'currency',
       currency: currency || 'EUR',
+      maximumFractionDigits: 6,
```
argaen, line 44: without this, it always returns at most 2 decimals

PR #844, opened 2024-04-09
Improve investment selection

### LINKED ISSUE #843: Demo and free versions are now showing incorrect investment type
Demo and free versions are showing investment parent account as ASSET type instead of INVESTMENT.

argaen:
I've changed how investments are selected in the placeholder page. Before only direct children were shown but now it selects any investments which `path` attribute starts with the `path` of the current account. This allows to have investment parent accounts within investment parents which makes it better for organisation.

Before this, you could not have a global investment parent with other parents within so you had to make that account an ASSET type.

Fixes #843


src/components/pages/account/InvestmentPlaceholderInfo.tsx
```diff
@@ -20,9 +20,10 @@ export type InvestmentPlaceholderInfoProps = {
 export default function InvestmentPlaceholderInfo({
   account,
 }: InvestmentPlaceholderInfoProps): JSX.Element {
-  const { data: investments, isPending } = useInvestments(
-    data => data.filter(d => account.childrenIds.includes(d.account.guid)),
+  const { data: investments = [], isPending } = useInvestments(
+    data => data.filter(d => d.account.path.startsWith(account.path)),
```
argaen, line 24: path.startsWith is a good way to check the investment is a leaf inside this account tree.

PR #845, opened 2024-04-19
Fix money to string decimals

argaen:
Currently, some moneyToString calls are showing too many decimals. This PR fixes that by allowing the user to specify how many decimals they want. This is necessary because when showing exchange rates, we need more than two decimals but for any normal money representation, 2 decimals is fine.

### DIFF
--- a/src/components/pages/commodity/PricesChart.tsx
+++ b/src/components/pages/commodity/PricesChart.tsx
@@ -42,7 +42,7 @@ export default function PricesChart({
           tooltip: {
             backgroundColor: '#323b44',
             callbacks: {
-              label: (ctx) => `${moneyToString(Number(ctx.parsed.y), currency)}`,
+              label: (ctx) => `${moneyToString(Number(ctx.parsed.y), currency, 6)}`,
             },
           },
         },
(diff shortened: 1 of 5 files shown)


src/components/pages/commodity/PricesChart.tsx
```diff
@@ -42,7 +42,7 @@ export default function PricesChart({
           tooltip: {
             backgroundColor: '#323b44',
             callbacks: {
-              label: (ctx) => `${moneyToString(Number(ctx.parsed.y), currency)}`,
+              label: (ctx) => `${moneyToString(Number(ctx.parsed.y), currency, 6)}`,
```
argaen, line 45: 6 decimals is arbitrary, can update if we find more are needed for some exchange rates.

src/helpers/number.ts
```diff
@@ -36,14 +36,16 @@ export function toFixed(n: number, decimals = 2): number {
   return parseFloat(n.toFixed(decimals));
 }
 
-export function moneyToString(n: number, currency: string): string {
+export function moneyToString(n: number, currency: string, decimals = 2): string {
   try {
     return n.toLocaleString(navigator.language, {
       style: 'currency',
       currency: currency || 'EUR',
-      maximumFractionDigits: 6,
```
argaen, line 44: yep, this broke the UI

PR #848, opened 2024-04-19
Remove free in favor of demo

argaen:
Having 3 different envs increases complexity too much when free and demo are almost the same. Changed the Demo storage instead to just return data or not depending on the window location.

With this change, im adding login to the demo website too. This increases consistency between prod and non prod to catch errors early in the login process.


src/lib/storage/DemoBookStorage.ts
```diff
@@ -10,6 +10,10 @@ export default class DemoBookStorage implements BookStorage {
   }
 
   async get(): Promise<Uint8Array> {
+    if (window.location.host === 'free.maffin.io') {
```
argaen, line 13: this is now the only difference between free and demo

PR #850, opened 2024-04-26
Basic SQL editor

### LINKED ISSUE #91: SQL editor
Implement a page that allows you to interact with the underlying data in the datasource. The SQL editor should be read/write so people can update their data "easily".

argaen:
SQL editor to interact with local data.

Screenshot 2024-04-26 at 12 49 29 PM

This is useful for fixing things on site or bulk edit stuff. It's also dangerous but the good thing is that the changes are not saved until you click save so. Note that if a user edits something with SQL without saving and then adds an account/tx it will save that.

### Future

This page has loads of potential as it could become something like databricks/etc where you can build graphs directly from the query results. Need to do some research to see if there is some syntax that can be used directly and tie it to chartjs. Another thing to add is autocomplete. Not a priority but it's a decent project to take on with external PRs

Closes #91

### DIFF
--- a/src/components/SQLExplorer.tsx
+++ b/src/components/SQLExplorer.tsx
@@ -1,0 +1,93 @@
+import React from 'react';
+import Table from '@/components/tables/Table';
+import { BiSolidDownArrow, BiSolidRightArrow } from 'react-icons/bi';
+import type { ColumnMetadata } from 'typeorm/metadata/ColumnMetadata';
+import type { PropsWithChildren } from 'react';
+
+import { useDataSource } from '@/hooks';
+import Loading from './Loading';
+
+export default function SQLExplorer(): JSX.Element {
+  const { datasource } = useDataSource();
+
+  if (!datasource) {
+    return <Loading />;
+  }
+
+  const metadatas = datasource.entityMetadatas;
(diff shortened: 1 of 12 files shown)


src/components/SQLExplorer.tsx
```diff
@@ -0,0 +1,93 @@
...
+              />
+            </Expandable>
+          </li>
+        ))
+      }
+    </ul>
+  );
+}
+
+function Expandable({
```
argaen, line 67: Probably can extract as a component to be re-used, will do whenever it is needed

src/layout/Menu.tsx
```diff
@@ -33,15 +34,14 @@ const MENU_ITEMS: ItemType[] = [
...
   {
-    label: 'Tools',
-    icon: <BiCog className="text-xl" />,
-    url: '',
-    submenu: [
-      {
-        label: 'Commodities',
-        url: '/dashboard/commodities',
-      },
-    ],
```
argaen, line 44: changed this coz couldn't find a quick way to show the menu to top direction and going bottom makes the menu to flicker (coz larger than the space left at the bottom of the page)

PR #856, opened 2024-05-12
Refactor transaction form pt.1

sseppola:
The end-goal with my contribution is to tame the TransactionForm which is quite difficult to work with, and my approach to do that is:
1. Centralize control of state in the 'root' of the form in order to have access to both all the events and state. By doing this I'm expecting to get rid of all `useEffects` because we can now update the entire state at once according to the event fired rather than relying on a downstream "sync" side effect based on the change in state.
2. Build a set of pure "operations" around the logic that is otherwise inlined in order to create a language around the intention of the expression. This will abstract business logic out of the components and into a layer independent of React, and forms the building blocks that we can use to compose bigger pieces of logic together. It is also very useful that logic is given a name so that it is easier to understand what's being done:
Screenshot 2024-05-12 at 13 50 24

This PR doesn't go all the way to refactor the TransactionForm because that's a rather big job and I would like to get your feedback on this sooner rather than later because it is quite different from existing style/architecture.

In this PR I attack the `SplitsField` because it is very central to the form but was also confusing to understand. I reworked it to make the logic clearer by delegating it to named operations, but also to make the component focused on its role of just "displaying and passing props to the fields". A key to this is introducing the `TransactionFormContext` domain object which is an abstraction that contain all state of the form so that we can build a set of "transaction form operations" which have all the context they need to be pure. By never "looking behind and breaking the abstraction" but instead using dedicated operations like:
```ts
const showMainSplit = txFormOps.hasMainSplitField(formContext);
const errorMessage = txFormOps.getSplitsErrorMessage(formContext);
```
it removes the components' coupling on the form library and the internal state of the form to give us leverage to change it over time without refactoring the components. In a following PR I'll double down on this structure and lift it into the `TransactionForm` and use it to pre-calculate how the form should render at the top level, which will allow me to intercept `onChange` calls made by the `MainField` in order to update the other visible form field.

Please let me know what you think of this, and if I can make it any simpler to review. I know it is a bit of a change in direction, but I'm open to discuss a middle ground here. 

One question on my part is that it was not clear to me how I could extend your entities with my "operations". I think it's neat to keep them separate as the "entity" is connected with the db, whereas the operations are supposed to be pure functions that give us the predictability that they do not mutate existing structures.

### DIFF
--- a/src/book/entities/priceOps.ts
+++ b/src/book/entities/priceOps.ts
@@ -1,0 +1,31 @@
+import type { DateTime } from 'luxon';
+import type { PriceDBMap } from '../prices';
+import type Commodity from './Commodity';
+import Price from './Price';
+
+type GetExchangeRateOptions = {
+  date: DateTime;
+  from: Commodity;
+  to: Commodity;
+};
+
+const getExchangeRate = (prices: PriceDBMap, options: GetExchangeRateOptions): Price => {
+  const { from, to, date } = options;
+
+  const isSameCommodity = from.guid === to.guid;
+  if (isSameCommodity) {
+    return Price.create({ valueNum: 1, valueDenom: 1 });
+  }
+
+  if (to.namespace !== 'CURRENCY') {
+    return Price.create({ valueNum: 1, valueDenom: 1 });
+  }
+
+  return from.namespace !== 'CURRENCY'
+    ? prices.getInvestmentPrice(from.mnemonic, date)
+    : prices.getPrice(from.mnemonic, to.mnemonic, date);
+};
+
+export const priceOps = {
+  getExchangeRate,
+};
--- a/src/book/entities/splitOps.ts
+++ b/src/book/entities/splitOps.ts
@@ -1,0 +1,7 @@
+import Split from './Split';
+
+export const splitOps = {
+  isSameCommodity: (a: Split, b: Split) => (
+    a.account?.commodity?.guid === b.account?.commodity?.guid
+  ),
+};
--- a/src/components/forms/transaction/SplitRow.tsx
+++ b/src/components/forms/transaction/SplitRow.tsx
@@ -1,0 +1,71 @@
+import React from 'react';
+import SplitField, { type SplitFieldProps, prepareSplitFieldProps } from './SplitField';
+import { TransactionFormContext, txFormOps } from './txFormOps';
+
+export type SplitRowProps = {
+  key: string;
+  splitFieldProps: SplitFieldProps;
+  deleteButtonProps?:
+  | { show: true; onClick: VoidFunction }
+  | { show: false };
+};
+
+export const prepareSplitRowProps = (
+  formContext: TransactionFormContext,
+  { index }: { index: number },
+): SplitRowProps | null => {
+  const showMainSplit = txFormOps.hasMainSplitField(formContext);
+  if (showMainSplit && index === 0) {
+    return null;
+  }
+
+  const { splitsFieldArray } = formContext;
+  const currentField = splitsFieldArray.fields[index];
+
+  const hasDeleteButton = index > 1;
+  const deleteButtonProps = hasDeleteButton
+    ? { show: true, onClick: () => splitsFieldArray.remove(index) }
+    : undefined;
+
+  return {
+    key: currentField.id,
+    deleteButtonProps,
+    splitFieldProps: prepareSplitFieldProps(formContext, index),
+  };
+};
+
+export function SplitRow(props: SplitRowProps) {
+  const { deleteButtonProps, splitFieldProps } = props;
+
+  return (
+    <fieldset className="grid grid-cols-12">
+      <div className="col-span-11">
+        <SplitField {...splitFieldProps} />
+      </div>
+      {(
+        deleteButtonProps?.show
+        && (
+          <div className="col-span-1">
+            <DeleteSplitButton {...deleteButtonProps} />
+          </div>
+        )
+      )}
+    </fieldset>
+  );
+}
+
+function DeleteSplitButton(props: {
+  onClick: () => void;
+}) {
+  return (
+    <div className="col-span-1">
+      <button
+        type="button"
+        className="btn btn-primary rounded-sm"
+        onClick={props.onClick}
+      >
+        X
+      </button>
+    </div>
+  );
+}
--- a/src/components/forms/transaction/txFormActions.ts
+++ b/src/components/forms/transaction/txFormActions.ts
@@ -1,0 +1,11 @@
+import { Split } from '@/book/entities';
+import { TransactionFormContext } from './txFormOps';
+
+const onAddSplit = ({ splitsFieldArray }: TransactionFormContext) => {
+  const newSplit = Split.create({ value: 0, quantity: 0 });
+  return splitsFieldArray.append(newSplit);
+};
+
+export const txFormActions = {
+  onAddSplit,
+};
--- a/src/components/forms/transaction/txFormOps.ts
+++ b/src/components/forms/transaction/txFormOps.ts
@@ -1,0 +1,32 @@
+import { UseFieldArrayReturn, UseFormReturn } from 'react-hook-form';
+import { FormValues } from './types';
+
+export type TransactionFormContext = {
+  action?: 'add' | 'update' | 'delete',
+  splitsFieldArray: UseFieldArrayReturn<FormValues, 'splits', 'id'>,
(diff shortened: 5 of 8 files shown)


src/components/forms/transaction/txFormOps.ts
```diff
@@ -0,0 +1,32 @@
...
+const getSplitsErrorMessage = ({ form }: TransactionFormContext) => (
+  form.formState.errors.splits?.message
+);
+
+export const txFormOps = {
+  getMainSplitValue,
+  getSplitsErrorMessage,
+  hasMainSplitField,
+  hasSplitFieldsOtherThanMain,
+};
```
argaen, line 32: I like the concept of the context for the form to encapsulate logic but I have the feeling we are duplicating something that comes already with the tooling. See https://react-hook-form.com/docs/useformcontext, maybe we could exploit that? I don't know if there is a way to customize the context to include this extra methods but maybe can do through extension? i.e. create a `useTxFormContext` that builds on top of that. Wdyt?

src/components/forms/transaction/txFormActions.ts
```diff
@@ -0,0 +1,11 @@
+import { Split } from '@/book/entities';
+import { TransactionFormContext } from './txFormOps';
+
+const onAddSplit = ({ splitsFieldArray }: TransactionFormContext) => {
+  const newSplit = Split.create({ value: 0, quantity: 0 });
+  return splitsFieldArray.append(newSplit);
+};
```
argaen, line 7: nit: I've been coding the whole project to use the `export function onAddSplit ...` can we keep it consistent? (if there is a pro/con over one or the other I'm happy to talk about it!)

argaen (commented):
I'm happy with the approach, anything that removes business logic from the UI components is a +1. Main thing is what I mentioned that maybe we can use the tooling provided by the form library

src/book/entities/splitOps.ts
```diff
@@ -0,0 +1,7 @@
+import Split from './Split';
+
+export const splitOps = {
+  isSameCommodity: (a: Split, b: Split) => (
+    a.account?.commodity?.guid === b.account?.commodity?.guid
+  ),
+};
```
argaen, line 7: maybe this could be part of the Commodity entity compare method and then have another in the Split as `sameCommodity`

src/book/entities/priceOps.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  return from.namespace !== 'CURRENCY'
+    ? prices.getInvestmentPrice(from.mnemonic, date)
+    : prices.getPrice(from.mnemonic, to.mnemonic, date);
+};
+
+export const priceOps = {
+  getExchangeRate,
+};
```
argaen, line 31: I'm thinking that maybe this one could be a method of the `PriceDBMap`. I've found myself repeating this logic in other places around the code where you have to check if it's investment or price or return a 1 so.. happy to move it there.

src/components/forms/transaction/SplitRow.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import SplitField, { type SplitFieldProps, prepareSplitFieldProps } from './SplitField';
+import { TransactionFormContext, txFormOps } from './txFormOps';
+
+export type SplitRowProps = {
+  key: string;
+  splitFieldProps: SplitFieldProps;
+  deleteButtonProps?:
+  | { show: true; onClick: VoidFunction }
```
argaen, line 9: lol didn't know about `VoidFunction` TIL

src/book/entities/splitOps.ts
sseppola: That makes a lot of sense, I assume I can just add functions to the classes?

src/components/forms/transaction/txFormActions.ts
sseppola: Sure, I can update that.

The idea is to have a namespace that communicates the domain of the function. Eg. `addOnSplit(foo)` could do anything, however `txFormActions.onAddSplit(foo)` is clear that it works with the transaction form.
It's possible to `import * as txFormActions from '...'` however that's optional and therefore not necessarily consistent, esp when auto import steps in to help

src/components/forms/transaction/txFormOps.ts
sseppola: That makes sense, however I plan on lifting the form context to the root of the form, so a custom hook won't be necessary as it's only used once.

The idea is to build an abstraction around `react-hook-form` so the components are not aware of which form library they're working with, but just communicate their intention by using the form operations (pure) and actions (side effects). If we wanted to switch out hook form for something else it should be possible to do mostly by working on these files.

While I don't intend to replace hook form I want to encapsulate it so I can pass our own `onChange` functions to have direct access to the form events. This will allow us to detect that the main field has been updated without relying to useEffects which just complicates data flow and fragments logic.

src/book/entities/splitOps.ts
sseppola: No 🙃 from logging it looks like in some contexts we just have plain json.

Screenshot 2024-05-19 at 22 29 28

Screenshot 2024-05-19 at 22 25 29

Also the Split type either has an error or exposes an error. It says a split always has an account, but you (very defensively) write: `splits[0].account?.commodity?.guid === splits[1].account?.commodity?.guid`
A split clearly doesn't always have an account because 8 tests fail with "TypeError: Cannot read properties of undefined (reading 'commodity')" I make that assumption. So is the problem in the type or the runtime?

src/components/forms/transaction/txFormOps.ts
argaen: okay, let's see the end result and then we can decide.

src/components/forms/transaction/txFormActions.ts
argaen: to me, where it is defined and how it is imported is enough, otherwise you find duplicating naming and stuff a lot for stuff that is not even intended to be reused outside its local scope. `forms/transaction` can be considered a local package and its functions local so IMO something like `forms/transaction/actions.ts` is something that contains actions for this package

src/book/entities/splitOps.ts
argaen: > No 🙃 when we get form values like form.getValues('splits.0.fk_account') as Account we get plain json instead of the class instance.

Ah... I know why: https://github.com/maffin-io/maffin-app/blob/master/src/components/forms/transaction/TransactionForm.tsx#L48

so... typically they are instances but not in the case of the splits. I'm a bit confused though coz with the `...split` the instance methods should also be forwarded? 🤔 

Anyway, I'm happy with 1) or 2)

> It says a split always has an account, but you (very defensively) write: splits[0].account?.commodity?.guid === splits[1].account?.commodity?.guid

The problem is that when creating a new transaction, you get partial state. As a developer we can pass `Partial<Split>` because we want to define some default values for the users automatically or also, just pass an empty split and then let the user fill everything. This is why the `splits[0].account?.commodity?.guid === splits[1].account?.commodity?.guid` is so defensive. At the end state, for sure, splits always MUST have an account but at the beginning, the user may not have selected an account yet. 

Not sure if I explained that well, happy to chat more about it and if there are better options :)

sseppola: > I'm a bit confused though coz with the ...split the instance methods should also be forwarded? 🤔

Not if that method comes from the prototype chain, `...` only copies properties without accessing `__prototype__` eg. "own properties". I don't know how TypeOrm works so I don't know where the boundaries are.

I suggest we make 1) a type `SplitJSON` that only contains the data properties, and 2) a function that is responsible for copying data from the instance to a new object because I doubt `...` is going to be enough. By having a function we have one place to change it if we find other issues.

> Anyway, I'm happy with 1) or 2)

Then I think we should do 2) because that allows us to group the functions that works on data in a without side effects and uses other entities' operations: `Split.operations.isSameCommodity(split1, split2)`
This can be very useful when debugging because "operations" keyword will tell you that this is pure, so you can rule out database interactions and direct instance mutations.

argaen: > I suggest we make 1) a type SplitJSON that only contains the data properties, and 2) a function that is responsible for copying data from the instance to a new object because I doubt ... is going to be enough. By having a function we have one place to change it if we find other issues.

Ok, let's go with the json type for Split and see how it works.

> Then I think we should do 2) because

maybe even the function to transform an instance to an object could be part of the operations too? But yeah sounds good to me :)

PR #860, opened 2024-05-15
Add account description field

### LINKED ISSUE #859: Display description field in account creation form
Some accounts need extended description so users can explain what's the intended purpose of that account (i.e. for specific taxes, if they are anual, etc).

argaen:
Fixes #859


src/book/entities/Account.ts
```diff
@@ -116,7 +116,6 @@ export default class Account extends BaseEntity {
     nullable: true,
   })
   @IsOptional()
-  @Length(4, 2048)
```
argaen, line 119: This length validators I just found them to be annoying and don't bring much value so removing

PR #862, opened 2024-05-15
Fix multiple split INCOME cash flow

### LINKED ISSUE #861: Wrong income calculation for multiple splits
In the detail page for bank accounts the total income calculation is wrong for a transaction with multiple incomes. ~I suspect the same may happen with the expenses.~ This seems to happen to any type of transaction with multiple splits

Try:

Bank +100
Income1 -75
Income2 -25

The total Income reflected is 200 instead of 100

This seems duplicate of #857 because also the sankey diagram is showing wrong quantities

argaen:
Fixes #861

### DIFF
--- a/src/__tests__/hooks/api/useCashFlow.test.tsx
+++ b/src/__tests__/hooks/api/useCashFlow.test.tsx
@@ -328,7 +328,7 @@ describe('useCashFlow', () => {
    *
    * This case is simple as we want two rows same as the splits.
    */
-  it('works with 3 splits in same currency', async () => {
+  it('works with 3 outflow splits in same currency', async () => {
     const account3 = await Account.create({
       guid: 'guid3',
       name: 'Expense2',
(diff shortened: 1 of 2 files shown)


src/__tests__/hooks/api/useCashFlow.test.tsx
```diff
@@ -389,7 +389,86 @@ describe('useCashFlow', () => {
...
 
   /**
-   * The above case is simple however, when
+   * Same as the test before but with receiving money into the ASSET
+   * account instead of withdrawing
+   *
+   * As an example, say we have a tx with the following:
+   *  - 100 in account1
+   *  - -75 in account2
+   *  - -25 in account3
```
argaen, line 398: Before the change this was returning

-100 account2
-100 account3

src/hooks/api/useCashFlow.ts
```diff
@@ -46,12 +46,14 @@ export function useCashFlow(
             ) AS mnemonic,
             SUM(
               CASE
-                WHEN splits.guid = credit_split.guid THEN
+                WHEN splits.guid = asset_split.guid THEN
```
argaen, line 49: credit_split was incorrect because for income transactions this means we are selecting the wrong account. Changed the logic to select asset type account (except INVESTMENT) which keeps all the tests passing and also passes the newly added one (plus correct behavior). Whenever we discover a new edge case can always update

PR #866, opened 2024-05-20
Fix negative asset values

### LINKED ISSUE #858: Negative amounts in assets not displaying
It is possible for bank accounts to go negative values. Currently the assumption is that it can't so even when the amounts are negative, it shows as positive which is confusing (probably an abs() somewhere). Should correct this to display correct information in the accounts tree and the account detail page

argaen:
Fixes #858 

For the Pie chart, it may be a bit confusing to users as we are displaying positive and negative values in the same dataset. Didn't find a better way to do this for now though so keeping as is:

Screenshot 2024-05-20 at 10 52 27 AM

### DIFF
--- a/src/components/pages/account/TotalWidget.tsx
+++ b/src/components/pages/account/TotalWidget.tsx
@@ -5,7 +5,6 @@ import Money from '@/book/Money';
 import type { Account } from '@/book/entities';
 import StatisticsWidget from '@/components/StatisticsWidget';
 import TotalChange from '@/components/widgets/TotalChange';
-import { isLiability } from '@/book/helpers';
 
 export type TotalWidgetProps = {
   account: Account,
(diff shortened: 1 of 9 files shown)


src/components/pages/account/TotalWidget.tsx
```diff
@@ -21,7 +20,7 @@ export default function TotalWidget({
     <StatisticsWidget
       className="mr-2"
       title="Total"
-      stats={isLiability(account) ? total0.format() : total0.abs().format()}
+      stats={account.type === 'INCOME' ? total0.abs().format() : total0.format()}
```
argaen, line 23: we show the real sign for all account types except for the INCOME one

src/lib/queries/getAccountsTotals.ts
```diff
@@ -33,7 +33,10 @@ export default async function getAccountsTotals(
   const totals: { [guid: string]: Money } = {};
 
   rows.forEach(row => {
-    totals[row.accountId] = new Money(row.total, accountsMap[row.accountId].commodity.mnemonic);
+    totals[row.accountId] = new Money(
+      accountsMap[row.accountId].type === 'INCOME' ? Math.abs(row.total) : row.total,
```
argaen, line 37: for INCOME accounts, make it a positive quantity so we avoid having to do Math.abs in each visualization

PR #907, opened 2024-05-24
Use nextjs actions for price fetching

argaen:
Sooo I was reading on adding authentication to external apis for nextjs with auth0 and discovered nextjs actions. This is a proposal to move price fetching to nextjs actions which:

- Allows us to remove stocker amplify backend completely which means only one code base, one pipeline, no cors, etc.
- Makes it easier to work with auth0 for api calls (I'm improving this because planning to try integrate with plaid).

Planning to deploy this and see how it behaves. Shouldn't be dangerous as price retrieval is not in the critical path and covered with global try/catch.

PS: If this works, I'll open another PR to remove all extra code and move the tests around

### DIFF
--- a/src/app/actions/getTodayPrices.ts
+++ b/src/app/actions/getTodayPrices.ts
@@ -1,0 +1,11 @@
+'use server';
+
+import getPrices from '@/lib/yahoo';
+import type { Price as PriceInfo } from '@/lib/yahoo';
+
+export default async function getTodayPrices(
+  tickers: string[],
+): Promise<{ [ticker: string]: PriceInfo }> {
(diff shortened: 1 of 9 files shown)


src/app/actions/getTodayPrices.ts
```diff
@@ -0,0 +1,11 @@
+'use server';
+
+import getPrices from '@/lib/yahoo';
+import type { Price as PriceInfo } from '@/lib/yahoo';
```
argaen, line 4: if one day we add an extra fetcher, we can always add an abstraction layer

PR #915, opened 2024-05-25
Plaid MVP

argaen:
MVP for Plaid that:

- Connects to sandbox to pull bank information
- Creates accounts returned by plaid into your DB

TODO:

- Access tokens created by plaid are unique and re-usable (they actually charge based on how many you have) so they should be stored somewhere. I'm thinking on adding a metadata object to the account.
- Deal with secrets.


src/lib/external/plaid.ts
```diff
@@ -0,0 +1,50 @@
+import { AccountType } from 'plaid';
+import type { TransactionsSyncResponse } from 'plaid';
+
+import { Account, Commodity } from '@/book/entities';
+
+export async function createEntitiesFromData(data: TransactionsSyncResponse) {
```
argaen, line 6: basically a DTO to transform from Plaid to our Entity models

src/app/actions/plaid.ts
```diff
@@ -0,0 +1,50 @@
...
+  PlaidEnvironments,
+  CountryCode,
+} from 'plaid';
+
+const client = new PlaidApi(new Configuration({
+  basePath: PlaidEnvironments.sandbox,
+  baseOptions: {
+    headers: {
+      'PLAID-CLIENT-ID': '6625de1ca0f038001c8a2e04',
+      'PLAID-SECRET': '<secret>',
```
argaen, line 16: This is a server action so code is executed in the backend

src/app/actions/plaid.ts
```diff
@@ -0,0 +1,59 @@
...
+ * https://plaid.com/docs/api/tokens/#linktokencreate
+ */
+export async function createLinkToken(userId: string) {
+  const response = await client.linkTokenCreate({
+    user: {
+      client_user_id: userId,
+    },
+    client_name: 'Maffin',
+    products: [Products.Transactions],
+    country_codes: [CountryCode.Es],
```
argaen, line 32: Need to either let user select or see if can just add all of them?

amplify.yml
```diff
@@ -8,6 +8,7 @@ frontend:
       commands:
         - rm -rf .next/
         - cp .env.$USER_BRANCH .env
+        - echo "PLAID_SECRET=$PLAID_SECRET" >> .env
```
argaen, line 11: PLAID_SECRET is an environment variable in AWS amplify. Tried with the secrets feature but has no support to fetch secrets locally unless you are running an amplify sandbox which means extra costs when developing

package.json
```diff
@@ -18,7 +18,6 @@
     "@tanstack/react-query-devtools": "^5.37.1",
     "@tanstack/react-table": "^8.17.3",
     "@testing-library/jest-dom": "^6.4.5",
-    "aws-amplify": "^5.3.12",
```
argaen, line 21: unrelated, just cleaning up

src/app/actions/index.ts
```diff
@@ -23,3 +24,16 @@ const wrappedGetTodayPrices = ({
...
+const wrapperCreateLinkToken = ({
+  userId,
+}: {
+  userId: string,
+}): ReturnType<typeof createLinkToken> => (
+  createLinkToken({
+    accessToken: Actions.accessToken,
+    userId,
+  })
+);
```
argaen, line 37: the wrapped functions can probably be generated dynamically, once we start having more, definitely something to look at

src/app/actions/plaid.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export async function createLinkToken({
+  userId,
+  accessToken,
+}: {
+  userId: string,
+  accessToken: string,
+}): Promise<string> {
+  await verify(accessToken);
+  if (!(await getRoles(accessToken)).isBeta) {
```
argaen, line 34: I'm only protecting the link token creation as without a link token, there is no moving forward in the link creation so this is enough.

PR #918, opened 2024-05-26
Inject access token to actions

argaen:
Pass access token to actions for better access control and permissions. 

Related to #915 so we can fine grained control.

This involved some changes in auth0 side:

- Create a role called 'premium' for paid users
- Create an api with audience `https://maffin`
- Create an action that includes the roles into the access token as a custom claim
```
   const namespace = 'https://maffin';

   if (event.authorization) {
      api.accessToken.setCustomClaim(`${namespace}/roles`, event.authorization.roles);
   }
```

This gives us a way to segment users within the same tenant in Auth0 using roles. This means now we can leave maffin-dev only for development and move all users to the production tenant.

### DIFF
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -32,6 +32,7 @@ const config = {
   moduleNameMapper: {
     '^@/(.*)$': '<rootDir>/src/$1',
     '^uuid$': 'uuid',
+    "^jose": 'jose',
   },
 }
 
--- a/src/app/actions/index.ts
+++ b/src/app/actions/index.ts
@@ -1,1 +1,25 @@
-export { default as getTodayPrices } from './getTodayPrices';
+import getTodayPrices from './getTodayPrices';
+
+export class Actions {
+  static _accessToken: string;
+
+  static get accessToken() {
+    return Actions._accessToken;
+  }
+
+  static set accessToken(token: string) {
+    Actions._accessToken = token;
+  }
+}
+
+const wrappedGetTodayPrices = ({
+  tickers,
+}: {
+  tickers: string[],
+}): ReturnType<typeof getTodayPrices> => (
+  getTodayPrices({
+    accessToken: Actions.accessToken,
+    tickers,
+  })
+);
+export { wrappedGetTodayPrices as getTodayPrices };
--- a/src/lib/auth0-provider.tsx
+++ b/src/lib/auth0-provider.tsx
@@ -20,7 +20,7 @@ export default function Provider({
         redirect_uri: (typeof window !== 'undefined' && window.location.origin) || '',
         scope: CONFIG.auth0.scopes,
         connection: 'maffin-gcp',
-        audience: 'https://premium',
+        audience: 'https://maffin',
       }}
     >
       <AccessTokenActionProvider>{children}</AccessTokenActionProvider>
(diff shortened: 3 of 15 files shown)


src/app/actions/index.ts
```diff
@@ -1 +1,25 @@
-export { default as getTodayPrices } from './getTodayPrices';
+import getTodayPrices from './getTodayPrices';
+
+export class Actions {
```
argaen, line 3: This is used as a global class to store the access token used for the actions. It's not clean and was thinking about using local storage but auth0 recommends to keep access tokens in memory.

src/lib/auth0-provider.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+function AccessTokenActionProvider({
+  children,
+}: React.PropsWithChildren): JSX.Element {
+  const { getAccessTokenSilently, isAuthenticated } = useAuth0();
+
+  React.useEffect(() => {
+    async function saveAccessToken() {
+      const accessToken = await getAccessTokenSilently();
+      Actions.accessToken = accessToken;
```
argaen, line 39: Here is where we inject the access token to the Actions class

jest.config.mjs
```diff
@@ -32,6 +32,7 @@ const config = {
   moduleNameMapper: {
     '^@/(.*)$': '<rootDir>/src/$1',
     '^uuid$': 'uuid',
+    "^jose": 'jose',
```
argaen, line 35: https://stackoverflow.com/questions/76608600/jest-tests-are-failing-because-of-an-unknown-unexpected-token-export

jwks-rsa brings triggers export problem when testing

src/lib/auth0-provider.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <Auth0Provider
+      domain={CONFIG.auth0.domain}
+      clientId={CONFIG.auth0.clientId}
+      authorizationParams={{
+        redirect_uri: (typeof window !== 'undefined' && window.location.origin) || '',
+        scope: CONFIG.auth0.scopes,
+        connection: 'maffin-gcp',
+        audience: 'https://maffin',
+      }}
```
argaen, line 24: Inject the config directly here as we have one Auth0Provider only

src/setupTests.ts
```diff
@@ -4,6 +4,7 @@
 // learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom';
 import crypto from 'crypto';
+import { TextEncoder, TextDecoder } from 'util';
```
argaen, line 7: Related to https://github.com/maffin-io/maffin-app/pull/918/files#r1615085420

src/lib/auth0-provider.tsx
```diff
@@ -0,0 +1,52 @@
...
+}: React.PropsWithChildren): JSX.Element {
+  return (
+    <Auth0Provider
+      domain={CONFIG.auth0.domain}
+      clientId={CONFIG.auth0.clientId}
+      authorizationParams={{
+        redirect_uri: (typeof window !== 'undefined' && window.location.origin) || '',
+        scope: CONFIG.auth0.scopes,
+        connection: 'maffin-gcp',
+        audience: 'https://maffin',
```
argaen, line 23: this is **needed** or otherwise auth0 returns an opaque jwt token which can't be decoded/verified.

PR #919, opened 2024-05-26
Use env files for config

argaen:


amplify.yml
```diff
@@ -7,6 +7,7 @@ frontend:
     build:
       commands:
         - rm -rf .next/
+        - cp .env.$USER_BRANCH .env.local
```
argaen, line 10: important. This overrides the `.env.local` file so it uses the config for the branch we deploy

.env
```diff
@@ -0,0 +1,3 @@
+NEXT_PUBLIC_AUTH0_DOMAIN=maffin-dev.eu.auth0.com
+NEXT_PUBLIC_AUTH0_CLIENT_ID=mMmnR4NbQOnim9B8QZfe9wfFuaKb8rwW
+NEXT_PUBLIC_AUTH0_SCOPES=profile email
```
argaen, line 3: This is a valid default file for configuring nextjs according to https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#environment-variable-load-order.

All the other config files are not "valid" but we copy them in the build to `.env` depending on the user branch

amplify.yml
```diff
@@ -7,6 +7,7 @@ frontend:
     build:
       commands:
         - rm -rf .next/
+        - cp .env.$USER_BRANCH .env
```
argaen, line 10: this. Without this prod would not work because the `.env` file contains development/staging variables

PR #920, opened 2024-05-26
Use roles for premium/standard ui

argaen:
Use `isPremium` flag coming from auth0 jwt token instead of IS_DEMO_PLAN and IS_PLAID_PLAN which depended on the environment being used. This means that we can now have both premium and non premium users in both environments.

This defeats the purpose of having `free.maffin.io` so I'll be deleting the domain and remove the `check_subscription` from app.maffin.io so anyone can login (new users will always be non premium as the role is assigned manually for now).

This behavior is nice because it allows to add extra roles for trying new features. I.e. could have a role called `beta` and then use an `isBeta` flag to enable those features for only those users.

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,3 @@
 NEXT_PUBLIC_AUTH0_DOMAIN=maffin-dev.eu.auth0.com
 NEXT_PUBLIC_AUTH0_CLIENT_ID=mMmnR4NbQOnim9B8QZfe9wfFuaKb8rwW
-NEXT_PUBLIC_AUTH0_SCOPES=profile email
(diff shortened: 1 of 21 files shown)


.env
```diff
@@ -1,3 +1,3 @@
 NEXT_PUBLIC_AUTH0_DOMAIN=maffin-dev.eu.auth0.com
 NEXT_PUBLIC_AUTH0_CLIENT_ID=mMmnR4NbQOnim9B8QZfe9wfFuaKb8rwW
-NEXT_PUBLIC_AUTH0_SCOPES=profile email
+NEXT_PUBLIC_AUTH0_SCOPES=profile email https://www.googleapis.com/auth/drive.file
```
argaen, line 3: regardless of the env now users can be premium so google drive permission is needed

PR #930, opened 2024-06-03
Add bank config entity

argaen:
Add new entity for holding online bank configuration. Probably will need to modify in the future as right now the online use case we have is for holding Plaid integration information

### DIFF
--- a/src/book/entities/BankConfig.ts
+++ b/src/book/entities/BankConfig.ts
@@ -1,0 +1,27 @@
+import {
+  Column,
+  Entity,
+  OneToMany,
+} from 'typeorm';
+import {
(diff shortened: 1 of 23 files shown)


src/book/entities/BankConfig.ts
```diff
@@ -0,0 +1,27 @@
...
+import type Account from './Account';
+
+@Entity('bank_config')
+export default class BankConfig extends BaseEntity {
+  @Column({
+    type: 'text',
+    length: 2048,
+  })
+  @IsString()
+    token!: string;
```
argaen, line 20: will contain the access token for the plaid Item

PR #949, opened 2024-06-12
CSS theming with variables

argaen:
Moved dark/light theme to CSS variables so now only one class is needed in Tailwind to give color to components regardless of the theme. This makes it easier in the future to add new themes as only a new set of variables needs to be defined


src/css/globals.css
```diff
@@ -1,8 +1,38 @@
...
+    --foreground: #94a3b8;
+
+    --primary: #0e7490b3;
+
+    --success: #16A34A;
+    --warning: #f97316b3;
+    --danger: #ef4444b3;
+    --info: #06b6d4b3;
+    --misc: #7c3aed;
+  }
```
argaen, line 30: to add a new theme, just add a new `.<name>` class and let the user select the theme with the ThemeButton. This makes it better to improve branding later with a more elaborated palette of colors

tailwind.config.ts
```diff
@@ -24,7 +24,21 @@ const config = {
       sans: ['InterVariable', ...defaultTheme.fontFamily.sans],
     },
     extend: {
+      gridTemplateColumns: {
+        '13': 'repeat(13, minmax(0, 1fr))'
+      },
```
argaen, line 29: Adding this back which got lost in a previous PR and makes some layout to break

PR #957, opened 2024-06-14
Responsive side bar

argaen:
LeftSideBar for mobile is now hidden and shown when the hamburger menu is clicked:

Screenshot 2024-06-14 at 5 20 59 PM

In the future we should improve to do couple of things:

- animation to slide from left to right.
- expand the menu so the text is shown together with the icon rather than just icons.


src/components/ui/Modal.tsx
```diff
@@ -0,0 +1,65 @@
...
+  triggerClassName?: string;
+  showClose?: boolean;
+} & React.PropsWithChildren;
+
+export interface ModalRef {
+  openModal: () => void;
+  closeModal: () => void;
+}
+
+function Modal({
```
argaen, line 16: new generic Modal component that just adds some common functionality we want for our Modal components like a X close button, expose the openModal/closeModal as ref, etc

src/layout/LeftSidebar.tsx
```diff
@@ -1,20 +1,35 @@
...
+          triggerContent={<BiMenu className="text-2xl" />}
+        >
+          <div className="pt-5">
+            <Menu />
+          </div>
+        </Modal>
+      </div>
 
-      <Menu />
+      <div className="hidden md:inline-block fixed bg-background-700 left-0 text-center h-screen pt-20 w-20 pb-5 z-10">
```
argaen, line 24: This is the desktop side bar which is hidden by default

src/layout/Menu.tsx
```diff
@@ -79,17 +79,17 @@ export default function DashboardMenu(): JSX.Element {
 function MenuItem({ item, className }: MenuItemProps): JSX.Element {
   return (
     <li
-      className="group h-12 text-slate-400 hover:bg-cyan-700 hover:text-white hover:w-64 hover:rounded-r-sm"
+      className="group h-12 text-slate-400 lg:hover:bg-cyan-700 lg:hover:text-white lg:hover:w-64 lg:hover:rounded-r-sm"
```
argaen, line 82: Don't show text on hover on small screens as it's weird. We should instead just show the whole text expanded

PR #1100, opened 2024-12-23
Fix delete button logic

### LINKED ISSUE #958: Don't let parent accounts with children to be deleted
Currently the delete button is being shown for parent accounts that have children. The action of deleting is failing with `FOREIGN KEY constraint failed` but this shouldn't even be shown.

The condition for showing the delete button is in https://github.com/maffin-io/maffin-app/blob/master/src/components/pages/account/Header.tsx#L28.

EDIT: Apparently this is also for non parent accounts. I see accounts with splits have the delete button enabled? Need to fix this

CarlosABVA:
Fixes #958.

About the edit note on the issue, I haven´t been able to replicate an account with splits that could be deleted


CarlosABVA:
> However, your code already improves current behavior so I'm happy to merge as long as we add a test and fix the pipeline! If you need a hand, happy to help.

Alrite I added a couple tests to cover the updated logic

